PR #21, opened 2022-06-05
update pointer to Agoric's captp: endo

dckc:


dckc:
@cwebber knock knock? @kriskowal ?

tsyesika:
This seems good!

PR #22, opened 2022-06-17
Update link to Agoric’s / Endo’s CapTP

kriskowal:
Fixes #21


zenhack:
This appears to be the same patch as #21; I think @dckc was just fishing for someone with merge rights to hit the button (sadly, not me).

kriskowal:
Oh, of course. And it seems I do not either.

Issue #15, opened 2021-10-04
Replacing deliver.rdr with op:listen?


zenhack:
I'm going to open some issues for things that have been discussed on IRC/matrix.

Full transcript for this one below, but I will try to summarize: Goblins has an extra `listen` message type that doesn't exist in E's CapTP or Cap'n Proto. As I understand it this message tells the receiver that it should send the actual data associated with a promise. I suggested that this message could actually just entirely replace the `rdr` field in `deliver` or `Call.sendResultsTo` in capnp, and provide a bit more flexibility.

Something not mentioned in the discussion, but, I think with this in place doing "fire and forget" messaging becomes fairly simple:
just send the `deliver` and immediately follow it with a `gc-answer` (with the caveat that capnproto's version of `gc-answer`, `Finish`, is allowed to cancel side effects, so we might want to fuss with that).

```
@zenhack:matrix.org : I don't see a listen op here: http://erights.org/elib/distrib/captp/index.html and I don't think any of capnp's messages are equivalent either; what exactly is op:listen?
cwebber : isd: you're right, and this is a consequence of taking a Lambda The Ultimate approach to this. Subscribing to a promise doesn't necessarily mean a miranda method. there was a reason this came up in the design, and I need to get it written down. Zarutian_HTC did a translation in his port to and from what Agoric expects
however I can't talk about this at the second
I will document it and put it on the tracker soon though
@zenhack:matrix.org : Cool, I look forward to reading about it.
@zenhack:matrix.org : Hazarding a guess: is it just an explicit operator for "please send me the result," in contrast to capnp's Call.sendResultsTo field? I can see how splitting that off into a separate message would allow a bit more flexibility; right now I think at least both go-capnp and haskell-capnp just always set sendResultsTo = caller, since the API makes it hard to know in advance when the results won't be needed.
cwebber : isd: yes, it's separate
yeah you got it
@zenhack:matrix.org : cool.
cwebber : oh
also
@zenhack:matrix.org : Yeah, and I can see why doing lambdas by default would make that particularly necessary.
cwebber : isd: messages sent with <- do have a "reply-to" field
however
it's for non-question promises
or rather non "return-style"
it's also because of promise pipelining
being able to send an arbitrary message to any object
er
to any promise, before its result appears
well, if you are doing it lambda-centric rather than method-centric
then you need to be able to distinguish between "I'm sending this because I want to subscribe" and "I'm sending this because I want to invoke it"
or even invoke what it eventually pints to
point
s
[...]

PR #42, opened 2023-04-20
Add draft CapTP spec

tsyesika:


draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,629 @@
...
+1.  MUST verify the `desc:handoff-receive` (see `desc:handoff-receive` section). If it is incorrect, abort the handoff; otherwise, continue.
+3.  If the gift has already been deposited return the gift; otherwise return the gift, when is received.
+
+Note: Gifts are per-session. This means that gifts deposited within the `Gifter <-> Exporter` session should only be available for withdrawal within the `Receiver <-> Exporter session`. The receiver should be the only party able to withdraw the gift left by the gifter. Implementers should ensure that the management of gifts adheres to this per-session requirement, preventing unauthorized access to gifts.
+
+# Operations
+
+## [`op:start-session`](#op-start-session)
+
+On a new connection a key-pair for this session should be generated. This key-pair should be an EdDSA key-pair with a SHA512 hash.
```
davexunit: Maybe a comma would help here. "On a new connection, a key-pair..."

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,629 @@
...
+
+-   `fetch`
+-   `deposit-gift`
+-   `withdraw-gift`
+
+#### `fetch` Method
+
+This method is used to fetch an object from the bootstrap object. To use it you need a `swiss-number` which is a Binary Data type. This swiss number should correspond an object which exists in this session. The result will be the object which corresponds to this `swiss-number` or an error if the object does not exist or a swiss number was not provided.
+
+Here is an example of how to use this method is:
```
davexunit, line 449: Maybe either "Here is an example of how to use this method" or "An example of how to use this method is:"

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,629 @@
...
+             args               ; sequence
+             answer-pos         ; positive integer | false
+             resolve-me-desc]>  ; desc:import-object | desc:import-promise
+```
+
+The `resolve-me-desc` is a `desc:import-object` or `desc:import-promise` which represents a reference to a local object which will be notified upon the resolution of the promise.
+
+### Sending
+
+When sending a message which expects a result, a promise should be generated on the side which is sending the `op:deliver-only`. This promise should be provided to the object sending this message. See the [promises section](#promises) for more information.
```
davexunit: Should this be `op:deliver` instead of `op:delivery-only`?

tsyesika:
Thanks for the review @davexunit, I've addressed your feedback.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+When this is received, the signature must be valid using the corresponding
+public key. If the signature is not valid, the operation should be aborted.
+
+## [`desc:handoff-give`](#desc-handoff-give)
+
+You should have an overview of handoffs before reading this section, please read
+the [Third Party Handoffs section](#third-party-handoffs) first.
+
+This certificate is provided by the gifter to the exporter, the exporter then
+create a `desc:handoff-receive` which includes this `desc:handoff-give`
```
davexunit, line 781: "then creates"

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+CapTP offers several valuable features, including:
+
+- Powerful promises and promise pipelining, allowing remote value usage before
+  resolution.
+- Error propagation across the network.
+- Distributed garbage collection.
+- Secure third-party handoffs, even when messages are published.
+
+CapTP is built upon the foundation of the actor model, where each actor is
+referred to as an object. An actor model is system where objects pass messages
```
davexunit, line 29: Should this be either "The actor model is a system where..." or "An actor model is a system where..."?

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+established and set up, and can be used to send and receive messages with remote
+objects.
+
+## Sending and receiving messages
+
+When an object sends a message to another object across a CapTP boundary, it
+either expects a reply or not. If the object expects a reply a
+[promise](#promises) is created to represent the reply. To do this, we send a
+[`op:deliver`](#op-deliver) message. If the message is delivered without
+importing an answer, this is an [`op:deliver-only`](#op-deliver-only) message.
```
davexunit, line 107: This paragraph uses "a `op:deliver`" and "an `op:deliver-only`". Should these be made consistent?

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+4.  Get back driving noise
+
+The other CapTP session, which is making the car factory, car, and finally car
+noise, is able deliver those messages immediately to the local objects on their
+side as they become fulfilled and send the car noise back at the end.
+
+
+# [Cryptography](#cryptography)
+
+Each CapTP session has a per session key pair which is used for signing certain
```
davexunit, line 235: Sometimes it is "key pair", like here. Other times it is "key-pair". Could one of those variants be used for consistency?

draft-specifications/CapTP Specification.md
zenhack: @davexunit, I believe these are two different messages, the latter of which ignores any result.

davexunit: Oh let me be more clear: It's the "a" and "an" part that is inconsistent to me.

zenhack:
Some general high-level/structural comments:

- I'm not sure who the target audience of this document is supposed to be; there's a lot of stuff in here that I feel like would be incomprehensible to someone who hadn't already grokked what what using Goblins or something like it was like, and needlessly verbose to someone who has.

  My intuition is that probably the spec should be written for an audience that already has an understanding of what CapTP does for them, but may not know how that works at the protocol level at all.
- The document floats back and forth between describing things that are protocol-level considerations and things that are really descriptions of a typical API that a library might provide. This is confusing, and more clearly marking and organizing these would go a long way to making this easier to digest.

Some of the lower-level descriptions are useful, but I think this needs a lot of organizational work.

I have an obvious bias, but if I were going to take a whack at it, I'd probably start off by stealing much of the language and structure from rpc.capnp; bits will need to be tweaked for the differences between the protocols, but the general shape of the protocol is similar enough that I think drawing heavily on that would probably save a lot of work. Feedback on that documentation has generally been extremely positive, and multiple implementations have been written when referencing it.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+- Using an out-of-band mechanism to connect to a remote session
+- Peforming a handoff to connect to a third session
+
+In either situation, a secure channel needs to be created to the remote session.
+This is out of scope for this specification but it's covered by the OCapN
+Netlayers specification. Once a secure channel is established, we MUST to
+perform the following in this order:
+
+1.  Create a per-session cryptography key pair (see
```
dckc, line 88: I wonder why a session key pair is in this layer.

The Agoric CapTP API doesn't use a session key pair.

cc @erights @kriskowal @michaelfig

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+promise is created for the value of the response. They can be chained together
+in what is called [Promise Pipelining](#promise-pipelining), whereby messages
+are sent to the promise which should be delivered if the promise resolves.
+
+## [Promise and Resolver Objects](#promise-objects)
+
+Promises work like regular objects in CapTP. Promises come as a pair, there is a
+"promise" object which represents a value and may eventually resolve to either a
+value or object reference or may break, either explicitly, implicitly through
+error handling, or via network partition.
```
gibson042, line 144:
```suggestion
"promise" object which represents a value and may eventually fulfill to either a
value or object reference, or may break by either explicit
instruction or implicit error propagation or network partition.
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+Messages can be sent to a promise as if it were the resolved object. These sent
+messages will be relayed to the eventual object if it is `fulfill`ed to one.
+There is also a "promise resolver" object which is used to provide the promise a
+value or break on an error. The resolver object has two methods:
```
gibson042, line 149:
```suggestion
Messages can be sent to a promise as if it were an object. These sent
messages will be relayed to the eventual fulfilled value if it is an object.
There is also a "promise resolver" object which is used to provide the promise a
fulfillment value or break reason. The resolver object has two methods:
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+- `fulfill`: Provide the promise with its resolved values (success case)
```
gibson042, line 151:
```suggestion
- `fulfill`: Provide the promise with its fulfillment values (success case)
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+### `fulfill`
+
+This method can take any number of arguments. These arguments can be promise
+pipelined on with the [`op:pick`](#op-pick) operator and should be delivered to
+any listeners of the promise.
+
+### `break`
+
+Break takes a single argument, which represents an error that occured. This
```
gibson042, line 162: This is a surprising asymmetry, which probably merits some explanation.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+## [Promise pipelining](#promise-pipelining)
+
+When an [`op:deliver`](#op-deliver) message is sent and a result is pending, it
+is often desirable to use this result. One way to do this would be to wait until
+the promise has been resolved and then send the next message to the result.
+While this approach can be taken within OCapN, promise pipelining is preferred.
+When a message is sent with an `answer-pos` specified, the promise on the remote
+session can then be referenced and messaged with the
+[`desc:answer`](#desc-answer) descriptor. When the result is ready, messages
+sent to this promise will have them delivered to the result.
```
gibson042, line 182:
```suggestion
[`desc:answer`](#desc-answer) descriptor. When a promise is fulfilled
to an object, messages sent to the promise will be delivered to that object.
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+If the answer contains multiple values, the [`op:pick`](#op-pick) operator
+MUST be used to choose the value to be pipelined on. This works by exporting the
+individual value in another answer position, which then can be used to message
+the promise. If multiple values are pipelined on, without the pick operation,
+the resulting promise MUST be broken with an error and not fulfilled.
```
gibson042, line 188: So there are protocol-level affordances for fulfilling a promise to multiple values (and for automatically breaking messages that do not target a specific member of such a collection)? Interesting.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+## [Signature](#signature)
+
+Signatures are formatted using gcrypt's s-expression format and the EdDSA
+signature scheme. The formatted signature s-expression follows this structure:
+
+```text
+(sig-val (eddsa (r ...) (s ...)))
+```
+
+Everything exception the `...` must be verbatim. The `...` is replaced by Binary
```
gibson042, line 265:
```suggestion
Everything except the `...` must be verbatim. The `...` is replaced by Binary
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+Data typed values representing the signature.
+
+# [Third Party Handoffs](#third-party-handoffs)
+
+Third party handoffs are used when a message that is sent within a CapTP session
+contains a reference to an object that has been imported from a different CapTP
+session. This ability to include remote references to objects is a crucial part
+of CapTP as it enables objects to use any references they hold in any context.
+
+The third party handoffs specified here ensures this can be done in a secure
```
gibson042, line 275:
```suggestion
The third party handoffs specified here ensure this can be done in a secure
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+Third party handoffs define three roles:
+
+-   **Gifter**: The session sharing their `desc:import-object` or
+    `desc:import-promise`
+-   **Receiver**: The session the `desc:import-object` or `desc:import-promise`
+    is being shared with.
+-   **Exporter**: The session exporting the `desc:import-object` or
+    `desc:import-promise`
+
+## Handoffs from the Gifter's prespective
```
gibson042, line 287:
```suggestion
## Handoffs from the Gifter's perspective
```

gibson042 (commented):
Thanks for the very thorough explanation!

draft-specifications/CapTP Specification.md
zenhack: The 3PH presentation from a couple meetings ago goes over how they're used, but I do think it might be good to add some abstraction boundaries; probably most of the crypto should be netlayer specific.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+valid signature of `acceptable-location`.
+
+## [`op:bootstrap`](#op-bootstrap)
+
+This operation requests the bootstrap object to be available at the specified
+`answer-position`. The message looks like:
+
+```text
+<op:bootstrap [answer-position    ; positive integer or 0
+               resolve-me-desc]>  ; desc:import-object | desc:import-promise
```
zarutian, line 427: resolve-me-desc here and in op:deliver seem to be redundant functionality of what op:listen does. See #15 for why.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+
+This operation delivers a message to an object with the expectation of a return
+value which should be installed at the location specified in `answer-pos`. The
+`op:deliver` message is:
+
+``` text
+<op:deliver [to-desc            ; desc:export
+             args               ; sequence
+             answer-pos         ; positive integer | false
+             resolve-me-desc]>  ; desc:import-object | desc:import-promise
```
zarutian, line 528: resolve-me-desc here and in op:bootstrap seem to be redundant functionality of what op:listen does. See #15 for why.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+#### `deposit-gift` Method
+
+The deposit gift method is used in conjunction with sending a [Third Party
+Handoff](#third-party-handoffs). This method is used to deposit a gift which has
+been sent to the bootstrap object. It has two arguments:
+
+1.  A gift ID that is positive integer.
+2.  A `desc:import-object` or `desc:import-promise` which has been exported
+    within the given CapTP session.
+
```
zarutian, line 468: missing an argument:
3. VatId of the recipiant of the gift.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+#### `withdraw-gift` Method
+
+This method is used to send the [`desc:handoff-receive`](#desc-handoff-receive)
+in order to receive a gift. It has one arguments:
+
+- The `desc:handoff-receive`
+
+This should have been sent with the `op:deliver` operation, the response the
+bootstrap object should give is the gift which was (or will be) deposited.
+
```
zarutian, line 487: This actually needs two arguments:
1. gift id of the gift
2. VatId of the giver

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,978 @@
...
+(capability) if (and only if) it possesses a reference to one. In other words,
+"if you don't have it, you can't use it."
+
+CapTP offers several valuable features, including:
+
+- Powerful promises and promise pipelining, allowing remote value usage before
+  resolution.
+- Error propagation across the network.
+- Distributed garbage collection.
+- Secure third-party handoffs, even when messages are published.
```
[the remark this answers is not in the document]
tsyesika: I've changed this to "Secure third-party handoffs, even when CapTP messages are not kept secret." hopefully that's more clear. It's meaning that the secure third party handoff feature is not dependent on the messages between the three sessions being kept secret. This should work just as well if these messages were public for all to see.

draft-specifications/CapTP Specification.md
tsyesika: Yeah. If the promise only represents one value you of course don't need to use `op:pick`, it's just for those promises which represent multiple values. It used to be that promises could only represent a single value, and so we often encoded multiple values in lists, which was fine, but it prevented promise pipelining and caused a lot of unpacking in our code. We solved this by adding this operator and having it so promises could represent multiple values.

draft-specifications/CapTP Specification.md
zarutian: In Spritely Goblins case here it would refer to the publiv key (or the cryptographics hash there of) of the endpont that is to get the gift.

@markm has the nifty phrase "My name is how you know it is me." which captures the essence of the idea of self authenticated designators.

zenhack (commented):
Sorry this took me so long to get through. I've tried to restrict myself to commenting on things that are about the *spec*, rather than critiquing the protocol itself; I certainly came across a number of substantive things that seem worth improving, but we'll get those as part of our ongoing discussions.

In addition to my in-line comments, I will also say I think the doc could use a description up front about how the doc is organized. I spent the first chunk of my time reading this saying "I wish there was just a reference for all the message types somewhere" only to eventually get to that at the end. The nature of a doc like this is that you tend to skip around, so when one section references a particular structure, it's good to know how to go look up what that is; the beginning of the doc should orient the reader so they can do this.

draft-specifications/CapTP Specification.md
tsyesika: I don't think this is needed or adds anything to the handoff process.

The `desc:handoff-give` and the `desc:handoff-receive` already contain enough information that ensures that only the exporter can pick up the gift left for them.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,980 @@
...
+
+Promises work like regular objects in CapTP. Promises come as a pair, there is a
+"promise" object which represents a value and may eventually resolve to either a
+value or object reference or may break, either explicitly, implicitly through
+error handling, or via network partition.
+
+Messages can be sent to a promise as if it were the resolved object. These sent
+messages will be relayed to the eventual object if it is `fulfill`ed to one.
+There is also a "promise resolver" object which is used to provide the promise a
+value or break on an error. The resolver object has two methods:
```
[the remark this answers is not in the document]
tsyesika: Goblins promises are built upon regular actors/objects and use op:deliver/op:deliver-only to function. This means that a promise is created via op:deliver (or in Goblins, you can also spawn them as you can with any other object, and share them within a message over the CapTP boundary). The promise object can have messages sent to it, and these will be forwarded to the object (or promise) it eventually resolves to, unless it breaks.

Goblins doesn't have a CapTP level op:resolve or anything similar; it simply has this resolver object which is spawned at the same time as the promise. When it's sent a message, it expects the first symbol to be 'fulfill' or 'break', followed by one or more values.

This is how promises work, both over CapTP and locally, and why this "promise API" documentation ends up in the specification. There are two such object APIs in the CapTP spec from Goblins' perspective: the bootstrap object with its three "methods", and the promise resolver with its two "methods".

draft-specifications/CapTP Specification.md
tsyesika: I actually hadn't seen #15 before so thank you for linking it. I think the concept of dropping the `resolve-me-desc` and then simplifying the `op:deliver` and `op:deliver-only` messages into a single `op:deliver` message and then relying on `op:listen`more is really interesting. I discussed it briefly with Christine and she thinks it's an interesting idea too.

I think this for now, this would be better addressed as a follow up PR if the group generally think this is a good idea (I think my initial impression is that I agree, this might be a nicer solution) we go ahead and make the change later.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,980 @@
...
+    for both in band and out of band usage.
+
+# CapTP Overview
+
+CapTP supports many features, as described in the [Introduction](#introduction)
+section. It can, however, be broken down into the life cycle of a CapTP session.
+A session is between two entities which are communicating via CapTP through a
+OCapN Netlayer's channel. A CapTP life cycle consists of the following stages:
+
+1. A session is established pairwise between machines.
```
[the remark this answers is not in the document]
tsyesika: I've re-written this paragraph a little and moved away from life cycle closer to what you're suggesting here, let me know if this is what you had in mind.

tsyesika:
Thanks for everyone's feedback so far. I've made some changes based on everyone's feedback. I'd like to propose we could merge this as a preliminary draft so we can begin opening up issues where we have disagreement and begin working on followup PRs as we come to consensus.

Just replying to one of Ian's comments here as Github for some reason didn't let me inline:
> I think reworking this can wait for separate PRs after merging, but I think having connection semantics cleanly separated from the netlayer is important. Note, in particular, capnp often gets used over local unix sockets, where layering crypto on top would be silly...

This not crypto which is used to encrypt all messages, it's used in just two places:

- Once when setting up the session in the the `op:start-session` operation.
- On each handoff as part of the handoff certificates (`desc:handoff-give` and `desc:handoff-receive`).

I'm suspecting this overhead is relatively minor and is integral to the design. I think if you still think this is an issue, maybe this is better as a follow up issue as it would involve changing how handoffs work in a substantial way that would be better as it's own change set. I think two user space programs on the same machine speaking over unix sockets may still wish to have this security though.

zenhack:
I'm fine with merging this and addressing outstanding issues separately.

tsyesika:
I'm thinking it'd be good if we could take a vote on merging this in tomorrow's meeting (@jar398) It'd be good if folks could take a look if you've not already got to it and let me know if you have blocking feedback.

jar398:
I must be missing something - what is the risk of merging? I don't see that having something in the repo implies any kind of endorsement; traditionally endorsements would be spelled out in the document itself. All we're doing is making the document easier to find, and that seems like a good thing.

dckc:
Agoric hasn't completed its review; we'd like more time.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+```
+
+The `wants-partial?` flag indicates if a "partial" update should be provided as
+the notification when the promise resolves. If `wants-partial?` is false and the
+promise resolves to another promise, a notification to the `listen-desc` is not
+sent. A notification is only ever sent to the `listen-desc` when the promise
+resolves to a non-promise value, or breaks. If `wants-partial?` is true, a
+notification on any resolution, including a promise. In the case it does notify
+on a promise the listen is considered complete, if future notifications are
+desired, further `op:listen` operations should be sent to the promises.
```
gibson042, line 662: Grammar suggestions:
```suggestion
The `wants-partial?` flag indicates if a "partial" update should be provided as
the notification when the promise resolves to another promise. If
`wants-partial?` is false, a notification is sent to the `listen-desc` only when
the promise resolves to a non-promise value or breaks (i.e., not when it
resolves to another promise). If `wants-partial?` is true, a notification is
sent for any resolution, including to another promise. Any notification is
considered to conclude the `op:listen` interaction, and if future notifications
are desired (e.g., after a partial notification) then further `op:listen`
operations should be sent.
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+When receiving this message, providing `to-desc` matches a promise exported to
+this session, a mechanism MUST be setup to fulfill or break the provided
+resolver when a resolution is available to the `listen-desc` object. If a
+resolution is already available, the resolver provided in `listen-desc` MUST be
+fulfilled or broken.
+
+## [`op:gc-export`](#op-gc-export)
+
+When a reference is given out to another CapTP session, the reference must be
+kept valid until the other session is done with it. This is achieved by
```
gibson042, line 683: "other session", or other party within a session?

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+
+```text
+<op:gc-answer answer-pos>  ; answer-pos: positive integer
+```
+
+# Descriptors
+
+Several operations (e.g. `desc:import-object` and `desc:export`) are describing
+importing and exporting objects. There had to be a choice if these actions
+should be described from the sender\'s or receiver's side, in this case we
```
gibson042, line 722: Markdown nit: this backslash should be unnecessary (at least for GitHub)
```suggestion
should be described from the sender's or receiver's side, in this case we
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+choose the receiver's side. This means if an object is exported from session A
+to session B, session A sends a `desc:import-object` as session A is describing
+it from B's perspective.
```
gibson042, line 725: **session**, or **party**?
```suggestion
choose the receiver's side. This means if an object is exported from Alice to
Bob, Alice sends a `desc:import-object` to describe it it from Bob's
perspective.
```

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+```text
+<desc:import-object position>  ; position: positive integer
+```
+
+
+## [`desc:import-promise`](#desc-import-promise)
+
+When a promise is exported over a CapTP boundary is it described with a
+`desc:import-promise` message. This message contains a positive integer which is
+unique to this CapTP session and refers to this specific promise.
```
gibson042, line 742: Unique to the **session**, or unique to the **party** _within_ the session?

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+```
+
+## [`desc:sig-envelope`](#desc-sig-envelope)
+
+This encapsulates a CapTP object and provides a signature. The signature is
+created on the binary data of the serialized CapTP object in the `signed` field.
+
+The process of generating this is:
+
+1.  Fully serialize to Syrup octets a CapTP object.
```
gibson042, line 781:
```suggestion
1.  Fully serialize a CapTP object to a sequence of Syrup octets.
```

gibson042 (commented):
More comments—a mixture of grammar/Markdown suggestions, requests for clarification (in some cases including section reorganization, in some cases consistent use of vocabulary such as **session** vs. **party**, and in others just miscellany), and bigger questions such as where the line is drawn between convention and requirement (particularly with respect to method identification and multi-valued promise resolution).

draft-specifications/CapTP Specification.md
dckc: I promoted this thread to an issue: #53.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+2. Communication occurs between objects.
+    -   Initial connectivity to objects is established by querying the bootstrap
+        object for access to known objects (via sturdyrefs or certificates).
+    -   Messages are exchanged between objects which hold references to each
+        other.
+    -   Promises may be created when messages are sent.
+    -   Messages may be pipelined to promises, queueing those messages to
+        eventually be delivered to their resolution.
+    -   Handoffs are initiated when sending a message with a reference to an
+        object on a third CapTP session.
```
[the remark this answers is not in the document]
tsyesika: I think third is correct here, however I think christine was also confused by my use of "third" at one of the earlier passes through she did so I'll change this to different wording. I wanted to be very explicit that it's a reference outside of the session in question so I've gone with slightly different wording than you proposed here:

> - Handoffs are initiated when sending a message with a reference to an object outside of the CapTP session.

draft-specifications/CapTP Specification.md
tsyesika: yeah I agree, I've used session in a number of places in ambiguous ways. To me a session consists of two parties speaking CapTP with one another, however I've used it to mean a single party within a session which is confusing. I've fixed this wording here.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+
+This covers the basic usage of CapTP. Another aspect that can occur when sending
+messages through CapTP is when an object sends a message that includes a
+reference to an object residing in a different session. To perform this while
+maintaining the security CapTP provides, we need to perform a [Third Party
+Handoff](#third-party-handoffs).
+
+It is conventional, but not required, that some objects may offer different
+behavior depending on a "method". The method is conventionally a symbol whose
+value is the name of the method/behavior. This symbol is the first argument in a
```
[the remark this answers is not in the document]
gibson042: #54

dckc (approved):
Agoric has completed its review now. Modulo the issues we've raised and any editorial stuff from @gibson042 , this looks OK to merge.

cc @erights

tsyesika:
I've merged this as per the discussion we had in our last meeting.

tsyesika:
Oh I think github had "helpfully" hidden a lot of his feedback, to me it looked like I had addressed it. I'll follow up with another PR which addresses the rest of his feedback, sorry for missing it!

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+
+# Promises
+
+Promises are a key part of CapTP. They are used to represent a value which is
+not yet known. Promises are created by sending an `op:deliver` message, where a
+promise is created for the value of the response. They can be chained together
+in what is called [Promise Pipelining](#promise-pipelining), whereby messages
+are sent to the promise which should be delivered if the promise resolves.
+
+## [Promise and Resolver Objects](#promise-objects)
```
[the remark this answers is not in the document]
tsyesika: Yeah. I had wanted a good way to provide links to sections within a document and couldn't find a standard markdown way of doing this. However I did see a few suggestions to make your headings anchor links which you can then refer to later in the document.

Another solution I saw was to "guess" what the slug would be that markdown readers/renders would give it, but I wasn't really keen on this solution (I suspect the slug could change depending markdown renderer).

I'm going to leave the headlines links for now as I've made heavy use of linking to sections within the rest of the document, however I'm open to other solutions (including potentially using a different markup format, if folks prefer).

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+Public keys are formatted based on gcrypt's s-expression format, using EdDSA
+public keys and the SHA512 hash algorithm. The EdDSA public keys are based on
+the Ed25519 elliptic curve. The public key is formatted as follows:
+
+```text
+(public-key (ecc (curve Ed25519) (flags eddsa) (q ...) (s ...)))
+```
+
+Everything except the `...` must be verbatim. The `...` is replaced by Binary
+Data typed values representing the public key.
```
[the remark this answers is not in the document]
tsyesika: This representation is based off the s-expression format that gcrypt uses, so this exact expression (except with `q` and `s` values filled with the public key data). The expression is actually a Syrup sequence composed of symbols. Spritely is not attached to this format and we probably would be looking to change it within the future.

I think this might be a good candidate for it's own `desc` struct, but I think this could be best discussed within the group to see what works best for everyone here.

draft-specifications/CapTP Specification.md
```diff
@@ -0,0 +1,1002 @@
...
+covered by the [`op:abort`](#op-abort) message. When this is received the
+session should be severed and unresolved promises broken.
+
+# Promises
+
+Promises are a key part of CapTP. They are used to represent a value which is
+not yet known. Promises are created by sending an `op:deliver` message, where a
+promise is created for the value of the response. They can be chained together
+in what is called [Promise Pipelining](#promise-pipelining), whereby messages
+are sent to the promise which should be delivered if the promise resolves.
```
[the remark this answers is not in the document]
tsyesika: I've changed this to fulfill and hopefully made it clearer with the language I've used here.

draft-specifications/CapTP Specification.md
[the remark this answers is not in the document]
tsyesika: I've improved the text describing the two cryptography values, hopefully it's a bit clearer now.

Issue #29, opened 2023-01-13
OCapN URIs / introduction by serialized reference


tsyesika:
When it comes to sharing a sturdyref or having a machine address, we need a way to express this in a standard way. There are several things to consider here:

- What format of the URI to use?
- Do we need to support hints?
- What kinds of URIs do we support? (probably at minimum: URIs to a machine & sturdyrefs, but what about certificates for example?)

To kick the ball rolling on this Goblins uses this format for machines:

`ocapn://<transport-address>.<transport>`

and for a sturdyref is:

`ocapn://<transport-address>.<transport>/s/<swiss-num>`

jar398:
I want to question the need for these things to be URIs. I used to advocate for a URI form for sparse capabilities but after pushback from independent sources that I found plausible I'm not so sure any more. The argument is basically that, regardless of how wrong it is, it is human nature to put URIs (those found in log files, bookmark lists, and so on) onto insecure channels. The concern is mainly server-side leaks when using generic web toolkits.

Rather than argue about principles (e.g. "just don't put them on insecure channels" which is what I kept saying), first I would like to know the intended use cases that require them to be URIs.

The alternative is just to make them some kind of string that we don't call a URI. That would force them onto different channels from URIs. It's no guarantee of privacy on its own but it should help.

dckc:
> `ocapn://<transport-address>.<transport>`

Seems to me that should be `<transport>://<transport-address>` or perhaps `ocapn-<transport>://<transport-address>` or `<transport>+ocapn://...`.

zarutian:
I rather lean towards `ocapn:<transport>:<machineOrVatLocatorForThatTransport>/` as the common prefix. Then after the first / we get choices such as `s/<swissNum>` for swissNum (read “password capability” if you are looking into the old literature) based sturdyrefs and many others for all kinds of cerificates based sturdyrefs.

```ebnf
ocapn-uri := “ocapn:” transport-scheme-identifier “:” <vatLocator> “/“ ocapn-uri-payload
transport-scheme-identifier := “tor” | “tls” | “websocket” | “webrtc” | <others>
vatLocator := <depends on the transport but must be self authenticating>
ocapn-uri-payload := swissNumPayload | macaroonPayload | ActiveCapCertPayload | UcanPayload | BisquitPayload | <others>
swissNumPayload := “s/“ swissNum
macaroonPayload := “mr/“ macaroon
ActiveCapCertPayload := “acc/“ ActiveCapCert
[...]

PR #73, opened 2023-07-10
Draft OCapN Locators specification

tsyesika:


draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+node.
+
+## [Syrup Serialization](#node-syrup-serialization)
+
+It's encoded as a record with the label `ocapn-node` and three arguments:
+
+```
+<ocapn-node designator  ; string
+            transport   ; symbol (cannot contain ".")
+            hints>      ; hashmap | false
```
dpwiz, line 63: Would it be possible to use an empty map here, without union-ing it a subset of booleans?

Otherwise, when an empty map is provided, in which sense it is distinct from `false`?

draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+This is a URI with the scheme `ocapn` followed by the designator, a `.` and then
+the transport name. If any  hints exist they're added as part of the query
+parameters, otherwise emitted. 
+
+Note that the designator permits `.` to be used within it, however the final `.`
+should designate the seperator  between the designator segment and the transport
+identifier.
+
```
dpwiz, line 96: Can the transport bit be moved into the schema part of an URI?

That way the semantics will precede designator, one can sort and prefix filter URIs by transport without having to parse the designator, and the generic rfc3986 facilities can be used.

- `ocapn+tor://hsdfhsdjkhflasdkhf.onion:12345?hints...`
- `ocapn+tcp://[fd7a:115c::9f1b]:12345?hints...`
- `ocapn+cadet://secret@V8XXK9GAN5ZJFRFQP8MQX3D83BZTSBQVHKWWD0JPE63Z821906EG`

draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+"/s/" suffix to denoate that it's a sturdyref, followed by the swiss number
+value. Any hints are placed at the end of the URI.
+
+The URI format is as follows:
+```
+;; Without hints (i.e. hints are set to false)
+ocapn://<designator>.<transport>/s/<swiss-num>
+
+;; With hints
+ocapn://<designator>.<transport>/s/<swiss-num>?hint1=value1&hint2=value2
```
dpwiz, line 134: Do the hints relate to the transport or to the ref?

If it is the latter, then the ref can be encoded as an URI `fragment`, leaving hierarchy part (authority+path) to the node:

`ocapn://<designator>.<transport>/?hint1=value1&hint2=value2#swiss-num`

dpwiz:
Given how URI encoding is already in the spec, why not just use it directly, without introducing custom wrapping?

`<ocapn-node "scheme:authority/path?query#fragment">`

draft-specifications/Locators.md
tsyesika: Yeah, this has actually been discussed a bit in https://github.com/ocapn/ocapn/issues/29 and when designing these URIs in Goblins one of the contenders we (spritely) were considering (even though we decided not to go with it).

draft-specifications/Locators.md
tsyesika: The hints are just additional information which might be needed to connect to the node. The onion netlayer doesn't need any but for example in Goblins there is a TCP+TLS netlayer which has the IP and port encoded in the hints (the designator is the hash of the TLS certificate).

One of the reasons `/s/swiss-num` was chosen was leaving the option to support many different locator types in the future (e.g. certificate locator types). This would then allow for `/c/...`  to be used and it'd be quite flexible.

dpwiz: It is possible (desirable, even) to have a type prefix in the fragment too (`ocapn://.....#sn:swiss-num`).

What irks me in `node/s/path?node-hints` is that a reference is injected in a middle of a node information. This prevents using of a simple tail split and potentially clashes with path (or path-like) information used by a node (see "http-over-unix-sockets" encoding problems).

ocdtrekkie:
Ian was particularly interested in a locator being able to provide more than one possible netlayers supported, I believe. Is it possible to encode multiple transports in this format?

tsyesika:
> Ian was particularly interested in a locator being able to provide more than one possible netlayers supported, I believe. Is it possible to encode multiple transports in this format?

This doesn't explicitly mention anything for this but one way this could be done is with a locator type which is specifying multiple locators e.g.

```
ocapn://<self-authenticating-designator>.multi?onion=<onion-locator-here>&libp2p=<libp2p-locator-here>&etc=<etc...>
```

draft-specifications/Locators.md
tsyesika: Yeah, I think it could be an empty map, I'm not sure which is better. I'm certainly open to the change though.

It's usage of `false` here is mostly due to its roots in scheme which uses false as a value when something is lacking a specific value, like a lot of languages use a void/null like type (syrup doesn't have a other null like type).

draft-specifications/Locators.md
tsyesika: Yeah this could be moved to the fragment section. Although I wonder if we should first decide in the group #74, if these things should be URIs at all. If they are I think this change could make sense.

draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+### Hints
+
+This is a hashmap of key and values which are designed to encode additional
+connection information that the  netlayer might need to reach the node. The keys
+should be symbols with the values as strings.
+
+There can be any number of hints, including none at all. If no hints are used
+this field should be set to false.
+
+## URI Serialization
```
dckc, line 75: How about adding a note that whether to have a URI Serialization at all is an open issue?

 - https://github.com/ocapn/ocapn/issues/29

It's actually 2 open issues, but I'd rather that #74 were closed as a dup of #29.

 - https://github.com/ocapn/ocapn/issues/74

draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+
+This identifies an OCapN node, not a specific object. This includes enough
+information to specify which netlayer and provide that netlayer with all of the
+information needed to create a  bidirectional channel to that node.
+
+The node locator include the following pieces of information (more details
+below):
+
+- **Designator**: Usually representing the key, however can be any value
+  determined by the netlayer
```
kriskowal, line 47: My understanding of @warner’s position is that the designator should be the fingerprint of a public key that is the same regardless of the transport, such that a nodes reachable over multiple transports have the same designator. That would suggest that the transport would be more tightly coupled to the hints.

A useful thought experiment: how would the locator appear for a node that is reachable via Tor and WebSocket over HTTPS, such that the client can choose the transport based on availability of transport clients from their node? My expectation is that the locator would be of the form `( fingerprint, nonce?, hints: ( transport, transport-specific-extra )+ )`.

draft-specifications/Locators.md
```diff
@@ -0,0 +1,137 @@
...
+designator and transport but  different hints and be considered to be the same
+node.
+
+## [Syrup Serialization](#node-syrup-serialization)
+
+It's encoded as a record with the label `ocapn-node` and three arguments:
+
+```
+<ocapn-node designator  ; string
+            transport   ; symbol (cannot contain ".")
```
kriskowal, line 62: This is another case where the distinction between symbol and string in JavaScript doesn’t correspond to the distinction in Scheme.

draft-specifications/Locators.md
kriskowal: I would move that the transport move the other direction, such that client can filter connection hints for supported transports. Concretely:

```
[ocapn]
key: b0b5c0ffeefacade
nonce: deadb33f
[hints]
- tor://hsdfhsdjkhflasdkhf.onion
- tcp://[fd7a:115c::9f1b]:12345
- cadet://secret@V8XXK9GAN5ZJFRFQP8MQX3D83BZTSBQVHKWWD0JPE63Z821906EG
```

dckc (commented):
I think the differences between this draft and Agoric's preferences/plans are represented in open issues, so I'm fine with this going out, subject to change as we address those issues.

tsyesika:
Based on the approval in https://github.com/ocapn/ocapn/pull/82, I'm going to merge this.

PR #84, opened 2023-09-15
Remove op:bootstrap and instead always export at `0`

tsyesika:
This removes the `op:bootstrap` operation which used to ask the remote node make available the bootstrap object and would setup a promise. That is almost always unnecessary as the bootstrap object is needed to do anything useful with CapTP and so must always be issued.

This instead removes the operation and just makes it required that after a successful `op:start-session` exchange the bootstrap object be exported at position `0` which simplifies the spec.

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -94,9 +94,7 @@ perform the following in this order:
 2.  Send a [`op:start-session`](#op-start-session) message
 3.  Receive and verify the remote party's
     [`op:start-session`](#op-start-session)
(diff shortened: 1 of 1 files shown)


jar398:
@tsyesika I hadn't realized this was a PR so isn't subject to "record consensus by closing"... I'll put this on the agenda for November

tsyesika:
We have [decided last meeting (November 2023)](https://github.com/ocapn/ocapn/pull/98/files#diff-32c218c628e9b3844012db5dde0a36825caf23d8b1369d6f04288224e0b37fd0R44) to merge these.

PR #96, opened 2023-10-13
Add implementation guide

tsyesika:


cwebber:
YAY!  I wonder why it's not playing the diagrams?  I guess GitHub doesn't show those inline or something

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+```
+<op:deliver-only    <desc:import-object 3>
+                    ['fulfill "beeeep!"]>
+```
+
+Oh cool! that all worked. Now she has promises though she needs to modify how her `op:abort` works, before it just terminated the session if she got one, but now she has promises she needs to ensure they break if the session ends. She sets up the promise creation code to add her promise to a set of unresolved promises that her `op:abort` handling code can access. When the resolver gets the resolution, it removes itself from the set of unresolved promises. She next makes it so that when an `op:abort` comes in, she iterates through the set sending the resolver objects a break with a network partition error. This should ensure promises are handled correctly.
+
+Promises might not only be fulfilled through, lets imagine that Alisha sents another message to move forward. Lets assume the robot can't do that, it doesn't really matter why, maybe the robot physically isn't able to move forward for there being an obstical, maybe it's got low power, or some other reason entirely, but the promise she gets back instead of being resolved as fulfilled instead gets broken:
```
zarutian, line 272:
```suggestion
Promises might not only be fulfilled through, lets imagine that Alisha sents another message to move forward. Lets assume the robot can't do that, it doesn't really matter why, maybe the robot physically isn't able to move forward for there being an obstacle, maybe it's got low power, or some other reason entirely, but the promise she gets back instead of being resolved as fulfilled instead gets broken:
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+
+```
+<op:gc-export export-pos   ; positive integer
+              wire-delta>  ; positive integer
+```
+
+The `export-pos` is the same position that's used for the import position and the wire-delta outstanding references to the object that's been received. OCapN's GC implementation is a collaborative process where by both sides keep track of the number of times a given object has been referenced in CapTP messages. When the importing side no longer needs the object, it should send an `op:gc-export` message, with the `wire-delta` being the outstanding reference count. Each time a reference to an object is received the CapTP session should keep track of this and can emit the `op:gc-export` with this amount and remove it from the count. The for keeping track of the wire-delta is the importing node could no longer need the reference and send a GC operation while at the same time the exporting side might send another message which refers to that object, if the exporting side were then to get this GC operation, the exporting side needs to know if it's safe to actually GC it or not.
```
zarutian, line 304:
```suggestion
The `export-pos` is the same position that's used for the import position and the wire-delta outstanding references to the object that's been received. OCapN's GC implementation is a collaborative process where by both sides keep track of the number of times a given object has been referenced in CapTP messages. When the importing side no longer needs the object, it should send an `op:gc-export` message, with the `wire-delta` being the outstanding reference count. Each time a reference to an object is received the CapTP session should keep track of this and can emit the `op:gc-export` with this amount and remove it from the count. The reason for keeping track of the wire-delta is the importing node could no longer need the reference and send a GC operation while at the same time the exporting side might send another message which refers to that object, if the exporting side were then to get this GC operation, the exporting side needs to know if it's safe to actually GC it or not.
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+- Answers: This is used to keep track of position to local promises for received `op:deliver` calls.
+
+Starting with questions Alisha modifies her function which creates the `op:deliver` message to add a unique integer for the `answer-pos` instead of the `false` she used earlier. She adds that integer to the questions table and returns her internal representation for this vow that she calls the remote promise descriptor. She also adds to her function the ability to pass in as the target of the message the remote promise descriptor and in that situation it'll check the answer table and send it to that position wrapped in the `desc:answer` record.
+
+Alisha modifies both her `op:deliver` and `op:deliver-only` message handling code to check if the `to-desc` field is referring to an object or an answer. In the case it's an answer she looks up in her answers table and finds the local promise object which she sends the messages to so they get deliverd when the promise is fulfilled. Alisha also adds to her `op:deliver` message handling code to check for if `answer-pos` is none-null and if so stores the position in her answers table along with the promise created for the `op:deliver`.
```
zarutian, line 352:
```suggestion
Alisha modifies both her `op:deliver` and `op:deliver-only` message handling code to check if the `to-desc` field is referring to an object or an answer. In the case it's an answer she looks up in her answers table and finds the local promise object which she sends the messages to so they get delivered when the promise is fulfilled. Alisha also adds to her `op:deliver` message handling code to check for if `answer-pos` is not null or `'false` and if so stores the position in her answers table along with the promise created for the `op:deliver`.
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+
+```
+<op:deliver-only <reference to Ben (desc:import-object)>  [<desc:handoff-give representing the reference to the gallery>]>
+```
+
+#### Depositing the gift
+
+![](./handoffs-3party2.webm)
+
+This is done by the gifter, in our example Alisha's node, Node A. Each node needs a per-session gifting table which stores a gift ID to some reference to an object or promise. The gift ID is just a positive integer (usually incrementing integer) which is unique to the session which represents the reference being shared. When a handoff is initiated it's to a specific object on a node, not just the Node, the gifting mechanism is providing a secure way to share a specific reference to the receiver. The way a node deposits a gift is on the bootstrap object which has a specific `deposit-gift` method.
```
zarutian, line 392: Which bootstrap object? The one Alisha got from Node C, the Exporter?

btw the `deposit-gift` method on the bootstrap object is nearly the same as the `provideFor()` on the NonceLocator in CapTP of E.

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+
+Alisha starts implementing this by creating a gift table for each session and an auto incrementing counter to track the gift IDs she's sending. Alisha also implements the `deposit-gift` method on her bootstrap object which takes two arguments a gift ID and a reference and she puts these things into her gift table she's just made. Alisha then starts by adding code to her `op:deliver` and `op:deliver-only` to look through all the arguments being sent and when it encounters a reference to an object initialize a handoff and replace that reference with a `desc:handoff-give`. When a handoff is initialized in Alisha's system, it gets the current next gift ID and sends a message to the deposit gift method on the bootstrap object, this looks like this:
```
zarutian, line 394:
```suggestion
Alisha starts implementing this by creating a gift table for each session and an auto incrementing counter to track the gift IDs she's sending. Alisha also implements the `deposit-gift` method on her bootstrap object which takes two arguments a gift ID and a reference and she puts these things into her gift table she's just made. Alisha then starts by adding code to her `op:deliver` and `op:deliver-only` to look through all the arguments being sent and when it encounters a reference to an object hosted neither by here node or the node at the captp connection it initializes a handoff and replace that reference with a `desc:handoff-give`. When a handoff is initialized in Alisha's system, it gets the current next gift ID and sends a message to the deposit gift method on the ?Exporter? bootstrap object, this looks like this:
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+
+```
+<desc:handoff-give receiver-key       ; the public key of the receiver in the gifter <-> receiver session
+                   exporter-location  ; OCapN Locator for the exporter
+                   session            ; The session ID for the gifter <-> exporter session
+                   gifter-side        ; the public key of the gifter in the gifter <-> exporter session
+                   gift-id>           ; Positive Integer or zero.
+```
+
+This includes a lot of information, some of which comes from the Gifter's session with the Receiver and some of it comes from the Gifter's session with the Exporter. It's important to keep in mind when thinking about handoffs is that it does not rely on keeping secrets, all messages sent can be public while remaining secure and handoffs are capabilities which only allow the receiver to obtain a reference to the object that the gifter deposits. The burdon for checking the information within the certificates rests with the exporter since it is the Node which has the reference to provide.
```
zarutian, line 415:
```suggestion
This includes a lot of information, some of which comes from the Gifter's session with the Receiver and some of it comes from the Gifter's session with the Exporter. It's important to keep in mind when thinking about handoffs is that it does not rely on keeping secrets, all messages sent can be public while remaining secure and handoffs are capabilities which only allow the receiver to obtain a reference to the object that the gifter deposits. The burden for checking the information within the certificates rests with the exporter since it is the Node which has the reference to provide.
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+                      handoff-count      ; positive integer
+                      signed-give>       ; desc:sig-envolope containing a desc:handoff-give received from the gifter
+```
+
+We can see the receiver adds some information, but most of it is provided already by the gifter and is included by including that signed certificate along with the `desc:handoff-receive` in the `signed-give` field. One piece of information that's included here is the `handoff-count`, each session has an inrecementing integer starting at zero which reflects how many handoff-receive's they've sent in the session. They also include an integer which reflects how many handoff receive's they've received (with them being the exporter) within the session. This is used to prevent replay attacks, the exporter must only provide the reference once to prevent replay attacks. The last thing to note here is this is also wrapped in a `desc:sig-envolope` with a signature, but this signature is created with the receiver's private key in the gifter <-> receiver session. Again, this might seem a bit strange for two reasons:
```
zarutian, line 432:
```suggestion
We can see the receiver adds some information, but most of it is provided already by the gifter and is included by including that signed certificate along with the `desc:handoff-receive` in the `signed-give` field. One piece of information that's included here is the `handoff-count`, each session has an inrecementing integer starting at zero which reflects how many handoff-receive's they've sent in the session. They also include an integer which reflects how many handoff receive's they've received (with them being the exporter) within the session. This is used to prevent replay attacks, the exporter must only provide the reference once to prevent replay attacks. The last thing to note here is this is also wrapped in a `desc:sig-envelope` with a signature, but this signature is created with the receiver's private key in the gifter <-> receiver session. Again, this might seem a bit strange for two reasons:
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+
+1. Again, we're sending the `desc:handoff-receive` to the exporter but using a key from another session, the gifter <-> receiver session.
+2. Since these keys are per-session and generated anew on each new connection, the exporter normally would have now way to verify this signature. The gifter, however, included the receiver's public key they used within the `handoff-give` (the `receiver-key` field) that they signed. The exporter is able to extract this key from the `desc:handoff-give` and use it to verify the signature on the `handoff-receive` (more on this later in the document).
```
zarutian, line 435:
```suggestion
2. Since these keys are per-session and generated anew on each new connection, the exporter normally would have no way to verify this signature. The gifter, however, included the receiver's public key they used within the `handoff-give` (the `receiver-key` field) that they signed. The exporter is able to extract this key from the `desc:handoff-give` and use it to verify the signature on the `handoff-receive` (more on this later in the document).
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+                    (ocapn-node "tcenolezzq7vleywviuvwl74dh2nhs3nf7lun5zuhtjpwhjed5ojw6qd" 'onion #f)
+                    <ID of the session Alisha and Carol's node have>
+                    <Alisha's public key in her session with Carol>
+                    5)
+                  <signature Alisha made with her private key in the session with Carol's Node>)))
+            ]
+           #f
+           <desc:import-object 1>>
+```
+/If en had had a connection to Carol's Node, Node C, his Node would just have used that instead of creating a new one/
```
zarutian, line 554:
```suggestion
/If Ben had had a connection to Carol's Node, Node C, his Node would just have used that instead of creating a new one/
```

implementation-guide/Implementation Guide.md
```diff
@@ -0,0 +1,632 @@
...
+;;;    Node 1.  Both nodes in this diagram are cooperating to preserve
+;;;    that Bob has access to Carol but that Carol does not have access to
+;;;    Bob, and that Carlos has access to Bob but Bob does not have access
+;;;    to Carlos.  (However there is no strict guarantee from either
+;;;    node's perspective that this is the case... generally it's in
+;;;    everyone's best interests to take a "principle of least authority"
+;;;    approach though so usually it is.)
+```
+
+The OCapN specifications leaves what happens after messages are delivered to the implementations and offers no opinions on what system might exist. OCapN has its roots in systems which implement an event loop called a *vat*. The vat has objects spawned in it and messages sent to these objects are queued up in a FIFO, the messages a delivered one at a time to the objects. Vats are transactional and if an error were to occur during message delivery, the transaction would be aborted leaving the vat in the prior state.
```
zarutian, line 628:
```suggestion
The OCapN specifications leaves what happens after messages are delivered to the implementations and offers no opinions on what system might exist. OCapN has its roots in systems which implement an event loop plus object heap called a *vat*. The vat has objects spawned in it and messages sent to these objects are queued up in a FIFO, the messages a delivered one at a time to the objects. Vats are transactional and if an error were to occur during message delivery, the transaction would be aborted leaving the vat in the prior state.
```

E Vats were not transactional, I think only WaterKen Vats were.

zarutian (changes requested):
Found a few typos, oprahs (missing words), and a semantic confusion re which bootstrap object is meant near the start of the 3ph section.

tsyesika:
Thanks for the feedback @zarutian, I've addressed those typos.

kriskowal:
In my role as the new Agoric ~Proconsul~ Editor, I have no objection to landing this. The text appears to be consistent with normative direction already established by the specification as written. Although we do not have consensus on every point, I’m content that issues are open or can be opened to redress those and that this is a valuable document to have in place and to edit into a consensus-consistent form going forward.

Please let me know if this is inconsistent with the groups expectations of editorial review. I would like very much to get to a point where the repository reflects consensus, but also have no desire to start over from a blank page to get there!

tsyesika:
> In my role as the new Agoric ~Proconsul~ Editor, I have no objection to landing this without normative review at our regular meetings. The text appears to be consistent with normative direction already established by the specification as written. Although we do not have consensus on every point, I’m content that issues are open or can be opened to redress those and that this is a valuable document to have in place and to edit into a consensus-consistent form going forward.
> 
> Please let me know if this is inconsistent with the groups expectations of editorial review. I would like very much to get to a point where the repository reflects consensus, but also have no desire to start over from a blank page to get there!

I agree this is an entirely non-normative document and it seems no one did have any objections, I'm going to hit merge, if anyone does feel like this was the wrong call we can always revert but I think it's had quite a bit of review now and again, is non-normative.

PR #100, opened 2023-11-29
spec: replace positive with non-negative for handoff-count

cab404:
Test suite uses 0 handoff count for its tests, and there are mentions of "0 or positive" in this document. For less ambiguity, changing that to "non-negative".


cab404:
I guess that applies to many usages of «positive» in the document.

kriskowal:
I agree that handoff count should allow for a zero quantity.

@tsyesika does the spec as written currently use 0 or 1 for the bootstrap object?

Agoric’s CapTP reflects a slightly different design concerning references into the four tables. We support a level of indirection. The in-band data refers to promises and objects by a slot index, and the envelope for the data is a compact array/list that maps the slot index to an arbitrary reference. For CapTP, those references are strings of the form `{p,o}{+,-}positive-number`. We cannot allow for zero because we use the sign to indicate whether the reference is in the near or far table.

I will trust @tsyesika’s judgement whether this change is technically consistent with material that has already landed and, provided that confirmation, approve this to land and discuss convergence on this design direction with a new issue: https://github.com/ocapn/ocapn/issues/101

tsyesika:
> @tsyesika does the spec as written currently use 0 or 1 for the bootstrap object?

It uses `0`.

As @jar398  mentioned in the meeting this PR is mostly correcting a mistake, currently 0 or positive is functionally how it works in Goblins and the OCapN test suite. This does seem interesting as a boarder case of "it's technically normative but it's also correcting a small wording mistake".

I think we agreed grey areas would be put to a vote so I'm thinking we probably should do that with this one. I don't think it should take long as it's a simple change.

erights:
Ping?

kriskowal:
I’ve proposed that we add this needs-consensus PR to next week’s OCapN meeting agenda. #164

davexunit:
This looks good!

PR #104, opened 2024-01-28
Propose TCP Testing Netlayer

cab404:
Born here: https://github.com/ocapn/ocapn-test-suite/pull/4/

As onion netlayer is a bit clunky for quick testing and iteration over implementations, I suggest TCP netlayer — a testing only netlayer. 

@tsyesika

### DIFF
--- a/draft-specifications/Netlayers.md
+++ b/draft-specifications/Netlayers.md
@@ -100,6 +100,22 @@ should be hosted on the port `9045` and using "ED25519-V3" for the key type.
 Upon creation of a hidden service tor provides the "Service-ID", this must be
 supplied as the "designator" in the OCapN Node Locator.
 
+# TCP Testing Netlayer
+
+This is a testing-only, simplistic netlayer, providing no security or anonymity
+whatsovever, and only intended to run tests without huge overhead of setting up
+Tor nodes.
+
+## Implementation
+
+When creating and decoding the OCapN locator, the following information is used:
+
+- **designator**: IPv4 address of a machine. Usually 127.0.0.1 for testing.
+- **transport**: The symbol `tcp`
+- **hints**: Hints are not used, so this may be set to false/omitted.
+
+Only send pure Syrup-encoded messages over the wire.
+
 # Funding
 
 This document has been written with funding through the [NGI Assure Fund](https://nlnet.nl/assure), a fund established by [NLnet](https://nlnet.nl) with financial support from the European Commission's [Next Generation Internet](https://ngi.eu) program. Learn more on the [NLnet project page]( https://nlnet.nl/project/SpritelyOCCapN#ack).
--- a/draft-specifications/Netlayers.md
+++ b/draft-specifications/Netlayers.md
@@ -103,7 +103,7 @@ supplied as the "designator" in the OCapN Node Locator.
 # TCP Testing Netlayer
 
 This is a testing-only, simplistic netlayer, providing no security or anonymity
-whatsovever, and only intended to run tests without huge overhead of setting up
+whatsoever, and only intended to run tests without the huge overhead of setting up
 Tor nodes.
 
 ## Implementation
--- a/draft-specifications/Netlayers.md
+++ b/draft-specifications/Netlayers.md
@@ -102,19 +102,26 @@ supplied as the "designator" in the OCapN Node Locator.
 
 # TCP Testing Netlayer
 
-This is a testing-only, simplistic netlayer, providing no security or anonymity
-whatsoever, and only intended to run tests without the huge overhead of setting up
-Tor nodes.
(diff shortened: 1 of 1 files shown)


draft-specifications/Netlayers.md
```diff
@@ -100,6 +100,22 @@ should be hosted on the port `9045` and using "ED25519-V3" for the key type.
...
+whatsoever, and only intended to run tests without the huge overhead of setting up
+Tor nodes.
+
+## Implementation
+
+When creating and decoding the OCapN locator, the following information is used:
+
+- **designator**: IPv4 address of a machine. Usually 127.0.0.1 for testing.
+- **transport**: The symbol `tcp`
+- **hints**: Hints are not used, so this may be set to false/omitted.
```
kriskowal, line 115: I believe this is consistent with the status quo, but I would like to entertain a different design. An OCapN locator needs to capture:

1. One transport-agnostic datum
2. One or more transport-specific data that can be _filtered_ according to their **transport**

I would suggest that the **transport** needs to (at least for now) be much more specific. TCP gives us the transport protocol, but does not specify the message framing protocol or CapTP version. If we want just one `tcp`, we will need a protocol negotiation handshake to allow for the evolution of the protocol, which I think we should avoid for performance reasons. I would prefer to capture all these dimensions in the **transport** datum. In this case, those dimensions are:

1. TCP for **sessions**
2. Syrup for **message framing** (Please confirm.)
3. Syrup **intermediate representation**.
4. OCapN (0) **CapTP protocol**.

By contrast, Agoric supports a number of combinations, but for example:

1. TCP for **session**
2. Netstring for **message frame**
3. JSON for **intermediate representation**
4. Endo/Agoric CapTP (0) for **CapTP**

And also:

1. WebSocket for **session** _and_ **message framing**
2. JSON for **intermediate representation**
3. Endo/Agoric CapTP 0 for **CapTP**

We also support domain sockets and named pipes for **session**, Chrome host messaging for **message framing**, structured clone in some cases instead of JSON, and of course intend to eventually add OCapN to the CapTP layer. We might even also bridge Syrup. We also have the compact-ordered as a candidate for the **intermediate representation** of message bodies. And, the CapTP layer also subsumes the separable concerns of the `{slots, body}` envelope.

To this end, I’m tentatively using exhaustive **transport** terms like `tcp+netstring+json+captp0` and would recommend one like `tcp+syrup+ocapn0` here. I do imagine that as we converge on a small set of good combinations of layers, that we will want to abbreviate to something like `tcp0`, `tcp1`.

To my eye, for TCP, both `tcp+syrup+ocapn0` and the IP address belong in an array (like, **hints**) and the designator should be a transport-agnostic sturdy-ref and whatever cryptography that entails. Tor muddies this with its onion address that _could_ serve as both sturdy-ref and transport-specific address, but I would hope it is transport-specific.

kriskowal (commented):
I’m enthusiastic about adding TCP for testing (and even as a bedrock for other protocols like Tor). I have reservations about the structure of OCapN locators in general.

In my role as editor on behalf Agoric, this is a normative change that should be discussed on our regular call.

In my role as an OCapN champion at Agoric, I would request that this proposal be landed but  led by a clarification to the specification that we do not yet have consensus on locator design in general and cite issue #29 and my comment in particular at https://github.com/ocapn/ocapn/issues/29#issuecomment-1631644218

draft-specifications/Netlayers.md
cab404: well, having proper schemas in protocol part would be incredibly nice, I agree — however I think that someone was against it (I would prefer having `ocapn+onion` and `ocapn+bson+gz+tcp`, instead of weird postfixes we have in the spec rn)

cab404: cc @dpwiz

cab404: and bear in mind this is outside of scope of this proposal)

cab404: and just from the top of my head, it will also present a challenge for negotiating supported protocols. e.g you can connect via `tcp` but `gz` is not implemented.

so yeah, that really does require a full new proposal)

tsyesika:
Since I ended up having to make some changes to how the TCP testing netlayer worked in the test suite, that had caused this PR to get stale. I've brought it back up to date with how it's used. I haven't incorporated @kriskowal feedback about versioning of the netlayer or other more granular info, as I think we've yet to decide on how to approach that in the group.

kriskowal (approved):
Happy to merge this but also want to encourage decoupling the message delimiter protocol from the message codec.

jar398:
This looks done to me, but since our process is to get go-ahead from both editors, I've added @tsyesika as a reviewer.

PR #105, opened 2024-02-01
Propose Addresses and References

kriskowal:
This expands on the design of [locators](https://github.com/ocapn/ocapn/blob/main/draft-specifications/Locators.md) to add support for protocol evolution and negotiation. This is intended to anchor a conversation about this alternative design. I have not woven it into the CapTP spec.

Locators issue https://github.com/ocapn/ocapn/issues/29
This elaborates on the idea I proposed https://github.com/ocapn/ocapn/issues/29#issuecomment-1631644218
Original locators PR https://github.com/ocapn/ocapn/pull/73
Subsumes TCP netlayer concern https://github.com/ocapn/ocapn/pull/104

### DIFF
--- a/draft-specifications/Addresses-References.md
+++ b/draft-specifications/Addresses-References.md
@@ -1,0 +1,222 @@
+# Warning: This is a draft specification likely to undergo significant change
+
+This draft specification has been written in anticipation of implementation
+by both Spritely Goblins and Agoric’s Endo Daemon and is a counter-proposal to
+[Locators](Locators.md).
+Over time this document will change, likely significantly as the group
+converges on the design of OCapN. If you're interested in being part of that
+work, please join!
+
+Authors: Jessica Tallon, Christine Lemmer-Webber, Kris Kowal, & The OCapN
+Pre-standardization Group.
+
+# [Introduction](#introduction)
+
+OCapN nodes can exist in a wide variety of network environments including
+servers, laptops, web pages, web extensions, and mobile apps.
+OCapN can connect any pair of nodes that can establish a session and exchange
+messages, though each node may support different network layers.
+For example, a web page might only be able to reach out with a WebSocket,
+which dictates TLS, a signed cerificate, and handles message framing.
+A mobile app might be able to reach out with a wide variety of transport and
+message framing protocols, but require a relay or peer to peer mesh to accept
+connections.
+Web pages and extensions can connect to each other with a MessagePort, and
+extensions can connect to local applications with host messaging.
+Host messaging brings its own message framing protocol.
+
+So, nodes must have the option of accepting connections over multiple
+network layers and, when requesting a connection, be able to filter addresses
+for mutually supported network layers.
+The variety of OCapN network layers must be sufficient to construct a graph
+from this variety of OCapN nodes and be able to evolve over time.
+
+# Additional Documents
+
+This document does not stand alone, it relies upon multiple other documents
+which together build up OCapN  (Object Capability Network) specifications.
+
+This specification uses the following other specifications:
+
+-   [Syrup](): The serialization format used to encode the in band
+    representation of the locator. 
+-   [Tagged Netstring](https://web.archive.org/web/20140210012056/http://tnetstrings.org/):
+    Specification for message framing.
+-   [OCapN Netlayers](): Specification to open a secure communication channel
+    between two sessions, often on different networks.
+-   [RFC3986: Uniform Resource Identifier (URI): Generic
+    Syntax](https://www.rfc-editor.org/rfc/rfc3986): For encoding URI
+    representations of locations.
+
+# [Abstract](#abstract)
+
+An OCapN **address** provides a node with all the information it needs to
+establish a session with another node.
+An address consists of the name of a network layer and transport-specific
+details.
+The name of the network layer captures multiple dimensions:
+
+- a **transport protocol** like TLS or Tor,
+- a **message framing protocol** like Netstring,
+- or, a combination of transport protocol and message framing protocol
+  like WebSocket over HTTP over TLS,
+- a **version** like CapTP Version 0
+
+An **reference** has an identifier and must have at least one address.
+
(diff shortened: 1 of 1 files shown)


draft-specifications/Addresses-References.md
```diff
@@ -0,0 +1,222 @@
...
+* Syrup OCapN Version 0 (draft)
+
+> We advise the TCP transport to listen specifically on IP `127.0.0.1` (rather
+> than the typical default of `0.0.0.0`) to make the node unreachable from
+> other hosts and avoid use on multi-tenant hosts.
+> For private connections between a user and a user-agent on a multi-tenant
+> system, relay OCapN through a UNIX domain socket or a Windows named pipe.
+
+The Syrup serialization is a record with the symbol `'tcp-syr0` followed
+by a TCP address, including the port.
```
dpwiz, line 104: Can we have port in a distinct field for JSON and syrup formats?
I'd rather parse a structure than run a sub-parser on a string.

draft-specifications/Addresses-References.md
```diff
@@ -0,0 +1,222 @@
...
+
+## TLS WebSocket with Syrup, Version 0
+
+The network layer named `wss-syr0` combines:
+
+* Signed Certificate TLS for sessions,
+* HTTP for transport,
+* WebSocket for message framing, and
+* Syrup OCapN Version 0 (draft)
+
```
dpwiz, line 163: Anything touching TLS should allow certificate pinning to facilitate using encryption without touching PKI.

The cert chain is a hairy thing on its own, but at least the basic "pin the root" has great feature-to-weight ratio.

kriskowal: This is not in my wheelhouse. I’d welcome suggestion text!

kriskowal:
I think this structure leaves OCapN room to grow into various client-side load balancing solutions

Prior art:
- gRPC Load Balancing https://grpc.io/blog/grpc-load-balancing/
- Google SRE Load Balancing in the Datacenter https://sre.google/sre-book/load-balancing-datacenter/
- Google SRE Load Balancing in the Frontend https://sre.google/sre-book/load-balancing-frontend/
- Google SRE Managing Load https://sre.google/workbook/managing-load/

I think we’re also pretty well set-up for flexible placement, given that a gateway could use 3-party-hand-off, effectively adding a level of indirection from gateway addresses to placement address.

- Orleans Architecture https://learn.microsoft.com/en-us/dotnet/orleans/overview

draft-specifications/Addresses-References.md
dpwiz: Basically there should be an array of CN/SHA256 pairs. In the simplest case of self-signed certificate it would be just one element with CN=hostname and hash of the certificate.

Something like `pin=hostname/NWZiNGJhMWE2NTFiYWU4MDU3ZWM2YjVjZGFmYzkzZmE3ZTBiN2Q5NDRkNmYwMmE0Yjc1MWRlNGUxNTQ2NGRlZiAqLQo=&pin=CA%20Root/M2Y0MGYxNTg4NjczMzcyODc0ZTVjOTI2NDc5NGE3NWM2ZmVhYzFlNDExNTI5YzcxMTEwYjZmOGNlZjlhMTdlOCAqLQo=` in URI form.
In JSON:
```json
{ ...
, "pins":
  [ {"cn": "hostname", "sha256": "NWZiNGJhMWE2NTFiYWU4MDU3ZWM2YjVjZGFmYzkzZmE3ZTBiN2Q5NDRkNmYwMmE0Yjc1MWRlNGUxNTQ2NGRlZiAqLQo="}
  , {"cn": "CA Root", "sha256": "M2Y0MGYxNTg4NjczMzcyODc0ZTVjOTI2NDc5NGE3NWM2ZmVhYzFlNDExNTI5YzcxMTEwYjZmOGNlZjlhMTdlOCAqLQo="}
  ]
}
```

draft-specifications/Addresses-References.md
```diff
@@ -0,0 +1,222 @@
...
+details.
+The name of the network layer captures multiple dimensions:
+
+- a **transport protocol** like TLS or Tor,
+- a **message framing protocol** like Netstring,
+- or, a combination of transport protocol and message framing protocol
+  like WebSocket over HTTP over TLS,
+- a **version** like CapTP Version 0
+
+An **reference** has an identifier and must have at least one address.
```
kriskowal, line 65:
```suggestion
An **reference** has an identifier and must have at least one address.
A serialized reference, particularly the URL serialization of a reference, is suitable for sending an invitation to connect to a node and _locate_ a value out-of-band.
```

kriskowal:
Notes from February 13 meeting:

This proposal needs to evolve to accommodate separate identifiers for the node (pubkey) and the value (swissnum).

This requires the explicit selection of a cryptography for a “meta netlayer” such that a single public key can straddle multiple transport protocols (e.g., Tor, Veilid, DH TLS, which have redundant authentication cryptography, but which still provide valuable confidentiality cryptography).

We should talk about ways to frustrate an attacker who presents a “reference” (in this document’s parlance) that includes in its connection hints a bytewise proxy that is consequently able to win the race to authenticate the node and then monitor or hinder timely delivery of subsequent messages. This kind of attack presupposes that the attacker wins the race to introduce a victim to a node.

We did not conclude whether to pursue this design direction or to achieve the same ends with a multi-route meta-locator based on the existing locators proposal.

We all agreed that URIs (a design embraced by both this proposal and the Locators spec) are icky. I believe they’re still invaluable since they touch QR codes, OS custom protocol hooks, automatic links in messages, and other out-of-band introduction mechanisms.

Not stated but assumed: supporting both protocol negotiation and adequate security properties may not be possible and it’s unclear what we would do instead. Perhaps we can at least peel off serialization versioning in hints by some other means.

kriskowal:
And @erights proposes some required viewing for reference https://www.youtube.com/watch?v=a0Wr2BvIixk&list=PLzDw4TTug5O0CTk1yiTb1b7-yTMTpoo0i

PR #125, opened 2024-07-09
Propose abstract passable data model

### LINKED ISSUE #92: Working document to capture current consensus on abstract syntax (data model)
#5 starting [here](https://github.com/ocapn/ocapn/issues/5#issuecomment-1549012122) (I'll call this "Ian's summary") proposes an abstract syntax (used synonymously with 'data model') for Ocapn.  Together with subsequent discussion in that issue and perhaps elsewhere, we may have consensus to a large extent on what the abstract syntax should be in the eventual Ocapn specification.

I would like for the current consensus on the abstract syntax to be captured in some living document we can use for reference in implementation, testing, and specification work. There will be fine points on which we do not yet have consensus, which can be called out in the document. It's fine if the document borrows significantly from Ian's summary and the rest of issue 5; it does not have to be 'publication quality'. Ideally at our November meeting we can call for consensus on this document.

Concrete syntax is out of scope here.  See #93

kriskowal:
Closes #92 by migrating (and lightly editing) the abstract data model consensus positions from the wiki into the draft specifications.

### DIFF
--- a/draft-specifications/Model.md
+++ b/draft-specifications/Model.md
@@ -1,0 +1,570 @@
+
+This document captures a summary of consensus and remaining contention for the
+OCapN data model and abstract syntax, excluding the concern of concrete
+representation of these on the wire, but including non-normative
+representations in a selection of implementation languages.
+
+Commentary in block quotes is not normative.
+
+> Non-normative commentary.
+
+# Value
+
+A value is any [Atom](#atom), [Container](#container),
+[Reference](#reference-capability), or [Error](#error).
+
+Atoms:
+
+- [Undefined](#undefined)
+- [Null](#null)
+- [Boolean](#boolean)
+- [Integer](#integer)
+- [Float64](#float64)
+- [String](#string)
+- [ByteArray](#bytearray)
+- [Selector](#selector)
+
+Containers:
+
+- [List](#list)
+- [Struct](#struct)
+- [Tagged](#tagged)
+
+References:
+
+- [Target](#target)
+- [Promise](#promise)
+
+# Atom
+
+Atoms are values that cannot contain or refer to other values.
+
+## Undefined
+
+A value representing the absence of a value.
+
+> - **Guile**: `*unspecified*`
+> - **JavaScript**: `undefined`
+> - **Python**: `None`
+
+If a JavaScript implementation of OCapN receives a [Struct](#struct) with an
+Undefined value for some key, the struct must be represented as an object
+that owns a property with the value `undefined` for that key.
+
+> Consequently, `JSON.stringify` for the same OCapN struct will omit the
+> property from the generated JSON object.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality),
+there is only one Undefined value and it is equal to itself.
+
+## Null
+
+([JSON](#json-invariants))
+
+A value representing `null` as distinct from `undefined` for the purpose
+of maintaining [JSON Invariants](#json-invariants).
+
+> - **Guile**: tentatively `json-null` (*imported*)
+> - **JavaScript** and **JSON**: `null`
+> - **Python**: `Null` (*imported*)
+
+If a JavaScript implementation of OCapN receives a [Struct](#struct) with
+a Null value for some key, the struct must be represented as an object
+that owns a property with the value `null` for that key.
+
+> Consequently, `JSON.stringify` for the same OCapN struct will have
+> a property with the value `null` in the generated JSON object.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality),
+there is only one Null value and it is equal to itself.
+
+## Boolean
+
+([JSON](#json-invariants))
+
+A value that is either true or false.
+
+> - **Guile**: `#f`, `#t`
+> - **JavaScript** and **JSON**: `false`, `true`
+> - **Python**: `False`, `True`
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality),
+the values True and False are equal only to their respective selves.
+
+## Integer
+
+An arbitrary precision signed integer.
+
+> - **Guile**: `-1`, `0`, `1`
+> - **JavaScript**: `-1n`, `0n`, `1n`
+> - **Python**: `-1`, `0`, `1`
+>
+> Note: We achieved consensus on the name `Integer` at the
+> [November 14, 2023 meeting]
+> (https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2023-11-14.md).
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), every
+Integer value is only equal to Integer values that represent the same
+arithmetic integer.
+
+## Float64
+
+([JSON](#json-invariants))
+
+An IEEE 754 64-bit floating point number.
+
+OCapN preserves the distinction between +0 and -0.
+
+| Language       | Negative | Positive |
+|----------------|---------:|---------:|
+| **Guile**      | `-0.0`   | `0.0`    |
+| **JavaScript** | `-0`     | `0`      |
+| **Python**     | `-0.0`   | `0.0`    |
+
+OCapN preserves positive and negative infinity.
+
+| Language       | Negative        | Positive       |
+|----------------|----------------:|---------------:|
+| **Guile**      | `-inf.0`        | `+inf.0`       |
+| **JavaScript** | `-Infinity`     | `Infinity`     |
+| **Python**     | `float('-inf')` | `float('inf')` |
+
+OCapN collapses the complete set of NaN values to a single abstract NaN.
+
+OCapN provides no support for other floating point precisions.
+
+> - **Guile**: `+nan.0`
+> - **JavaScript**: `NaN`
+> - **Python**: `float('nan')`
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/58
+>
+> OCapN round-trips all floating point numbers representable within IEEE 754
+> binary64, except that OCapN considers all NaNs as equivalent, that is, as jointly
+> representing a single abstract NaN value.
+> So, any concrete NaN representation may validly round trip even if it results
+> in a different concrete representation.
+> However, we encourage use of a canonical representation for NaN.
+>
+> Concretely, the canonical NaN is `0x7ff8_0000_0000_0000`, though this is not
+> a concern of the abstract syntax and data model.
+>
+> All real and finite double precision floating point numbers participate in the
+> JSON subset of OCapN.
+> We expect OCapN-compatible JSON codecs, including the JavaScript `JSON` codec,
+> to round-trip all numbers except NaNs, infinities, and negative zero, but all
+> other numbers expressible with an IEEE 754 double-precision float to survive a
+> round trip without loss of precision.
+> We also do not expect integers expressed with higher precision in JSON to
+> survive a round-trip through an OCapN-compatible JSON codec.
+>
+> Consensus on preserving -0:
+> - As of May 16, @erights insisted that -0 round trip to 0 https://github.com/ocapn/ocapn/issues/5#issuecomment-1550020450
+> - On May 23, @zenhack proposed preserving -0 https://github.com/ocapn/ocapn/issues/5#issuecomment-1560116857
+> - We converged out of band and Agoric has committed to preserving -0 https://github.com/endojs/endo/issues/1602
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality):
+- Every finite Float64 is equal to any other Float64 with the same bitwise
+  representation.
+- Every infinite Float64 is equal to any other infinite Float64 of the same
+  sign.
+- There is only one NaN value and it is equal to itself.
+
+> JavaScript's `Object.is` is consistent with Pass Invariant Equality, whereas
+> `==` and `===` are not.
+
+## String
+
+([JSON](#json-invariants)†)
+
+A sequence of Unicode code points excluding surrogates (U+D800-U+DFFF).
+Strings are distinguished from [Selectors](#selector) by type, not content.
+
+> - **Guile**: `""`
+> - **JavaScript**: `''`
+> - **Python**: `''`
+>
+> †Strings participate in the [JSON subset](#json-invariants) of OCapN except
+> any strings that contain surrogate code points.
+>
+> A string's content must be expressible in UTF-8.
+> Some two-byte Unicode encodings, as in JavaScript strings, can contain
+> 16-bit surrogate code _units_ in the range from 0xD800-0xDFFF.
+> Pairs of surrogate code units correspond to a single Unicode code _point_
+> greater than or equal to U+10000 and can be expressed in UTF-8.
+> However unpaired or lone surrogates have no valid expression in any UTF
+> and so cannot be carried by OCapN.
+>
+> Notes: [January 2024 meeting
+> notes](https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2024-01-09.md)
+> record that we agreed that strings can only be well-formed Unicode, that is,
+> cannot contain unpaired surrogate code points.
+> For JavaScript, if a string does not pass [the `isWellFormed`
+> predicate](https://github.com/tc39/proposal-is-usv-string), then it is not a
+> Passable string.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), a pair of
+Strings are equal if they have the same quantity of Unicode code points and
+have the same respective Unicode code points in order.
+
+## Selector
+
+A sequence of Unicode code points excluding surrogates (U+D800-U+DFFF).
+Selectors are distinguished from [String](#string)s by type, not content.
+
+> - **Guile**: symbols `'name`
+> - **JavaScript**: an object with two own properties:
+>   for the registered symbol key `passStyle`, the value is the string
+>   `selector`; and
+>   for the well-known symbol `toStringTag`, the value is a string consisting
+>   of the code points of the selector.
+>   ```js
+>   ({
+>     [Symbol.for('passStyle')]: 'selector',
+>     [Symbol.toStringTag]: 'name',
+>   })
+>   ```
+> - **Python**: `Selector('name')` where `Selector` is imported from `ocapn`.
+>
+> A selector's content must be expressible in UTF-8.
+> Some two-byte Unicode encodings, as in JavaScript strings, can contain
+> 16-bit surrogate code _units_ in the range from 0xD800-0xDFFF.
+> Pairs of surrogate code units correspond to a single Unicode code _point_
+> greater than or equal to U+10000 and can be expressed in UTF-8.
+> However unpaired or lone surrogates have no valid expression in any UTF
+> and so cannot be carried by OCapN.
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/46
+>
+> OCapN uses the name Selector to avoid the implication that they will
+> correspond to a language's symbol type in all languages that have a symbol
+> type.
+> Selectors may correspond to symbols in languages where a symbol is eligible
+> for garbage collection when there are no extant references.
+> At this time, JavaScript cannot safely use registered symbols like
+> `Symbol.for('name')` for OCapN selectors, because some implementations intern
+> registered symbols without possibility of eventual garbage collection.
+>
+> OCapN supports one operator for delivering both function application and
+> method invocation.
+> By convention, method invocation is equivalent to function application, where
+> the first argument is a selector followed by the remaining arguments.
+>
+> However, like symbols in Guile, selectors are values and can appear anywhere
+> values appear, including any argument position, inside a container, or as a
+> promise fulfillment value or rejection reason.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), a pair of
+Selectors are equal if they have the same quantity of Unicode code points and
+have the same respective Unicode code points in order.
+
+## ByteArray
+
+An array of 8-bit bytes.
+
+> - **Guile**: `#vu8()`
+> - **JavaScript**: `new ArrayBuffer()`
+> - **Python**: `b''`
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/48
+>
+> At the [November 14, 2023]
+> (https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2023-11-14.md)
+> meeting, we agreed to settle on the prefix "byte" and agreed to decide at the
+> next meeting based on a [reactji poll]
+> (https://github.com/ocapn/ocapn/issues/48#issuecomment-1811097196).
+>
+> [January 2024 meeting
+> notes](https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2024-01-09.md)
+> record that we agreed on ByteArray because it was the winner of the poll, and
+> we had already agreed to resolve this specific issue by poll among these
+> three choices.
+>
+> The JavaScript representation of a ByteArray is an `ArrayBuffer` which may
+> be made immutable with the proposed JavaScript [Immutable
+> ArrayBuffer](https://github.com/tc39/proposal-immutable-arraybuffer) feature.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), a pair of
+ByteArrays are equal if they have the same quantity of bytes and have the same
+respective bytes in order.
+
+# Container
+
+A container is a value that contains other values.
+
+## List
+
+([JSON](#json-invariants))
+
+A list of any quantity of values.
+
+> - **Guile**: `'()`
+> - **JavaScript** and **JSON**: `[]`
+> - **Python**: `()` (We have not discussed whether to use tuple or list. Tuple
+>   is less mutable, though enforcing immutability in Python is not likely to be
+>   a goal the way it is in hardened JavaScript.)
+>
+> Lists participate in the JSON subset of OCapN.
+>
+> We achieved consensus to name the type "List" at the [November 14, 2023]
+> (https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2023-11-14.md)
+> meeting.
+
+A pair of lists are equal or Equal for purposes of [Pass Invariant
+Equality](#pass-invariant-equality) if they are the same length and every
+respective value is equal, transitively.
+
+## Struct
+
+([JSON](#json-invariants))
+
+> The name "struct" is tentative.
+> https://github.com/ocapn/ocapn/pull/125
+
+A collection of unordered (key, value) pairs.
+Each key must be a [String](#string), and must be
+non-[Equal](#pass-invariant-equality) to any other key within a Struct.
+[Values](#value) within a Struct may be of heterogeneous type.
+
+> - **Guile**: `make-tbd-hash` a hash of undecided type
+> - **JavaScript**: `{}`
+> - **Python**: `{}`
+>
+> For the purposes of surviving a round trip, the order of appearance of
+> entries in the struct must not be important for determining equivalence.
+> A struct representation concretely using one key order may validly
+> round-trip into a struct representation using another key order.
+> However, we encourage concrete representations to use some canonical key
+> order, though this is not a concern of the abstract syntax and data model.
+>
+> A JavaScript object that owns any symbol-keyed properties or uses a prototype
+> other than `Object.prototype` cannot be passed as an OCapN struct.
+> The key `Symbol.for('passStyle')` is special and indicates the kind of OCapN
+> value the object represents.
+
+A pair of structs are Equal for purposes of [Pass Invariant
+Equality](#pass-invariant-equality) if they mutually posses a value for every
+key in the other struct and every respective value is equal to their own,
+transitively.
+
+> A pair of structs may be Equal regardless of the order of appearance of
+> fields.
+
+## Tagged
+
+A pair consisting of a tag and a [Value](#value).
+The tag must be a [String](#string).
+
+> OCapN provides a small number of container types with minimal semantics.
+> Tags allow protocols to emerge from values annotated to indicate a richer
+> variety of semantics.
+> For example, an arbitrary-precision decimal number might be represented as a
+> Tagged with a tag of "decimal" and a value that is a decimal
+> [String](#string) such as "3.14" (or alternatively, a value that is a
+> two-element [List](#list) of significand and decimal exponent
+> [Integers](#integer) such as [314, -2]).
+>
+> - **Guile**: currently a Guile record labeled `'(make-tagged 'tagName 10)` (**imported**)
+> - **JavaScript**: an object with three own properties:
+>   for the registered symbol key `passStyle`, the value is the string `tagged`;
+>   for the well-known symbol key `toStringTag`, the value is the tag;
+>   for the string key `payload`, the value is the tagged value.
+>   ```js
+>   ({
+>     [Symbol.for('passStyle')]: 'tagged',
+>     [Symbol.toStringTag]: 'tagName',
+>     payload
+>   })
+>   ```
+> - **Python**: `Tagged('tagName', value)` where `tag` is a `string` and `Tagged` is
+>   imported from `ocapn`.
+>
+> OCapN tagged data have a single value (are not variadic) [November 14, 2023]
+> (https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2023-11-14.md)
+>
+> What are tagged values for? https://github.com/ocapn/ocapn/issues/52
+
+Tagged values are equal or Equal for the purposes of [Pass Invariant
+Equality](#pass-invariant-equality) if the tag and value are Equal,
+transitively.
+
+# Reference (Capability)
+
+A value that can receive messages, either a Target or Promise.
+
+> References are the Capabilities in the name OCapN ("Object Capability
+> Network") that distinguish a Capability Transfer Protocol from mere RPC
+> ("Remote Procedure Calls").
+> References forward messages and support
+> [Promise Pipelining](https://en.wikipedia.org/wiki/Futures_and_promises#Promise_pipelining).
+
+## Target
+
+Targets represent either a local value or a remote value.
+OCapN can export references to local targets and can import references to
+remote targets.
+A local target handles deliveries and produces either a return value
+(fulfillment) or thrown error (rejection reason) for a message delivery.
+A remote target (a presence) forwards messages to its corresponding local target.
+
+> - **Guile**: a procedure
+> - **JavaScript**: to be proposed
+> - **Python**: to be proposed
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/49
+
+Targets have [Pass Invariant Equality](#pass-invariant-equality).
+A target might be sent from a local peer to a remote peer, then the remote peer
+may send that target back to the local peer.
+The sent target will be equal to the received target and no other value.
+
+## Promise
+
+A Promise represents the eventual return value (fulfillment) or thrown error
+(rejection reason) for a message delivery.
+A Promise is pending until settled with either a fulfillment or rejection
+reason.
+OCapN queues messages delivered to a Promise.
+If the eventual resolution of a Promise is another Promise, OCapN forwards
+the queued messages to the next Promise.
+If the eventual fulfillment value of the Promise is a Target, OCapN forwards
+the queued messages to the Target.
+OCapN does not forward messages to non-references (non-capabilities).
+
+> - **Guile**: to be proposed
+> - **JavaScript**: a JavaScript promise
+> - **Python**: to be proposed
+>
+> Tracking https://github.com/ocapn/ocapn/issues/55
+
+OCapN does not maintain [Pass Invariant Equality](#pass-invariant-equality) for
+promises.
+
+As with all other types, OCapN does maintain [Pass Type Invariance]
+(#pass-type-invariant) such that a local promise sent to a a remote peer
+arrives as a promise, regardless of whether it is pending or settled locally.
+
+OCapN maintains that, if a local promise is sent to a remote peer and then
+returned, either the fulfillment value or the rejection reason of the sent and
+received promises will satisfy the pass invariants applicable to their type.
+
+> For example, if the sent promise settles with a fulfillment value that is a
+> Target, the sent and received targets will be equal, because Targets
+> maintain [Pass Invariant Equality](#pass-invariant-equality).
+>
+> For another example, if the sent promise settles with a fulfillment value
+> that is a Struct, the sent and received structs will be equal, because all
+> [Containers](#container) maintain [Pass Invariant
+> Equality](#pass-invariant-equality).
+
+# Error
+
+A value capturing the reason for rejecting a delivery.
+
+> - **Guile**: to be proposed
+> - **JavaScript**: a JavaScript Error object
+> - **Python**: to be proposed
+>
+> We have not yet converged on consensus for any particular details about the
+> modeling of errors. The purpose of errors is typically to indicate that some
+> requested operation failed. The purpose of the contents of errors is to
+> preserve and convey diagnostic information, mostly to help debug problems,
+> such as the root cause of a surprising failure. This is a best-efforts
+> obligation, for which we have not yet decided either what contents are
+> required, nor what is allowed, nor what must be preserved as errors are
+> passed from one site to another. Until these details are decided, the only
+> hard requirement is that an error round trip to an error. We avoid any
+> interpretation for now as to whether it is the "same" error.
+>
+> https://github.com/ocapn/ocapn/issues/142
+
+# Pass Invariant
+
+> Any property that holds for corresponding values both locally and remotely
+> between a pair of OCapN peers during a session is a Pass Invariant.
+
+## Pass Type Invariant
+
+All values passable between OCapN peers have a single, invariant type.
+A value sent from one peer and received in another will have the same OCapN
+type in both peers.
+For any value sent from a local peer to a remote peer then returned to the
+local peer, the sent and received values will have the same type.
+
+## Pass Invariant Equality
+
+OCapN defines a flavor of equality, nominally Equality.
+Two values may be Equal under Equality.
+Values of different OCapN types are not Equal.
+Values of the same OCapN type may be Equal under conditions specific to their
+shared type.
+All OCapN types maintain pass invariant Equality except Promise and
+(tentatively) Error.
+
+All OCapN implementations, regardless of language, must model all OCapN types
+such that Equality for the in-language representation of values is consistent
+with the Equality and Pass Invariance of Equality for their corresponding OCapN
+model type.
+
+A type holds Equality invariant over passage between OCapN peers.
+Any pair of local values that are Equal can be passed to another peer and the
+respective remote values will also be Equal.
+
+It follows that, for any local value that is sent over OCapN, if that value
+returns to the local peer through any path through the network of remote peers,
+the sent and received values will be Equal.
+
+> OCapN Equal values may be locally distinguishable by other operators
+> depending on the language implementation and type.
+
+> In JavaScript, none of `==`, `===`, or `Object.is` are sufficient to compare
+> Equality of values, which instead must take into account the OCapN type for
+> each value.
+> For example, Float64 can be compared by `Object.is`, which preserves the
+> distinction between 0 and -0 and lack of distinction between any two NaN
+> values.
+> `Object.is` is also sufficient for Target provided that the OCapN
+> implementation holds invariant a single object identity for every Target.
+> The JavaScript ArrayBuffer representation of an OCapN ByteArray must be
+> compared byte-for-byte.
+> Also, `Object.is` is not sufficient for Container because container
+> equality is based on recursive equality of the contents.
+
+# JSON Invariants
+
+OCapN holds invariant that:
+
+- Any JSON text that a JavaScript peer can produce using
+  `JSON.stringify`,
+- excluding those in which the Unicode representation of any object member name
+  or string includes a surrogate code point,
+- can then be read back with `JSON.parse`,
+- sent to any other OCapN peer,
+- returned from any OCapN peer,
+- then saved back to an equivalent JSON text with `JSON.stringify`.
+
+> For this reason, OCapN supports a distinct [Undefined](#undefined)
+> and [Null](#null).
+> If OCapN coerced `null` to `undefined`, a JSON text like `{"key": null}`
+> would become `{"key": undefined}` passing through an OCapN network, and
+> consequently reduce to `{}` when returned to JSON format with
+> `JSON.stringify`.
+> If OCapn coerced `undefined` to `null`, an idiomatic JavaScript options
+> object like `{"key": undefined}` would return through the OCapN network as
+> `{"key": null}` and would `key` would no longer pass `key === undefined`
+> checks.
+> If OcapN refused to pass documents with either `null` or `undefined`, OCapN
+> would have too great an impedence mismatch with either idiomatic JSON
+> documents or idiomatic use of JavaScript objects.
+>
+> OCapN maintains that -0 and 0 are distinct [Pass Invariant](#pass-invariant)
+> values, whereas JavaScript's `JSON.stringify` renders -0 as `0`.
+> This is consistent with OCapN's JSON Invariants because all pass-invariant
+> JSON texts are in the range of `JSON.parse(JSON.stringify(x))`, which
+> cannot express -0.
+> This holds without regard for -0 being in the range of `JSON.parse`,
+> including `JSON.parse('-0')`, because of the narrowing behavior of
+> `JSON.stringify`.
+>
+> OCapN does not hold invariant the preservation of arbitrary JSON texts
+> through arbitrary JSON implementations in arbitrary languages.


draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+> - **Guile**: `*unspecified*`
+> - **JavaScript**: `undefined`
+> - **Python**: `None`
+
+## Null (JSON)
+
+A value representing `null` as distinct from `undefined` in JSON.
+
+> In order for the JSON subset of JavaScript to be passable, Undefined must be
+> distinguishable from Null.
```
gibson042, line 26: Hmm, this reads strangely. Isn't JSON inherently passable?

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+
+Comparing the identities of remotable values is pass-invariant, meaning that
+comparing the identities of any pair of remotable values will produce the same
+result on one peer if passed over OCapN to another peer and compared there,
+regardless of whether either value is a presence on either peer.
+
+Although Remotables and Promises are both Capabilities for delivering messages,
+Promises do not support pass-invariant identity comparison.
+
+> - **Guile**: a function
```
davexunit, line 280: s/function/procedure/ but idk a callable actor is not a core concept in Guile, it's a Goblins thing.

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Promise
+
+A promise represents the eventual return value (fulfillment) or thrown error
+(rejection reason) for a message delivery.
+
+Messages delivered to a promise are forwarded to their eventual fulfillment
+value or rejected with the promises's eventual rejection reason.
+
+> - **Guile**: to be proposed
```
davexunit, line 303: There's no promise API in Guile. It's a Goblins addition.

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+>   }
+>   ```
+> - **Python**: `Selector('name')` where `Selector` is imported from `ocapn`.
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/46
+>
+> OCapN uses the name selector to avoid the implication that they will
+> correspond to a language's symbol type in all languages that have a symbol
+> type.
+> Selectors may correspond to symbols in languages where a symbol is elligible
```
gibson042, line 187:
```suggestion
> Selectors may correspond to symbols in languages where a symbol is eligible
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+> - **JavaScript**: a JavaScript promise
+> - **Python**: to be proposed
+>
+> Tracking https://github.com/ocapn/ocapn/issues/55
+
+## Error
+
+A value capturing the reason for rejecting a delivery.
+
+> - **Guile**: to be proposed
```
davexunit, line 313: This would be an exception object of some kind.

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+> with UTF-8, which cannot transit lone/unpaired surrogates.
+> (October 10, 2023)
+
+## ByteArray
+
+An array of bytes.
+
+> - **Guile**: `#vu8()`
+> - **JavaScript**: `new ArrayBuffer()`
+> - **Python**: `b''`
```
davexunit, line 151: The Guile and Python examples show literals but the JS one is a mutable thing. In Guile, the equivalent to `new ArrayBuffer()` to make a mutable bytevector would be `(make-bytevector len)`.

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+> round-trip into a struct representation using another key order.
+> However, we encourage concrete representations to use some canonical key
+> order, though this is not a concern of the abstract syntax and data model.
+>
+> A JavaScript object that owns any non-string keys (numbers or symbols) or
+> with a prototype other than `Object.prototype cannot be passed as an OCapN
+> struct as all such objects are reserved for concrete representations of OCapN
+> types.
+> The key `Symbol.for('passStyle')` is special and indicates the kind of OCapN
+> value the object represents.
```
gibson042, line 234:
```suggestion
> A JavaScript object that owns any symbol-keyed properties or with a prototype
> other than `Object.prototype` cannot be passed as an OCapN struct, as all such
> objects are reserved for concrete representations of OCapN types.
> The key `Symbol.for('passStyle')` is special and indicates the kind of OCapN
> value represented by an object.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/49
+
+## Promise
+
+A promise represents the eventual return value (fulfillment) or thrown error
+(rejection reason) for a message delivery.
+
+Messages delivered to a promise are forwarded to their eventual fulfillment
+value or rejected with the promises's eventual rejection reason.
```
gibson042, line 301:
```suggestion
> - **Guile**: a function
> - **JavaScript**: A *far object* or a *remote presence* of that far object.
>   ```js
>   {
>     __proto__: {
>       [Symbol.for('passStyle')]: 'remotable',
>       [Symbol.toStringTag]: allegedInterfaceName,
>     },
>     ...methods
>   }
>   ```
> - **Python**: to be proposed
>
> Tracking: https://github.com/ocapn/ocapn/issues/49

## Promise

A promise represents the eventual return value (fulfillment) or thrown error
(rejection reason) for a message delivery.

Messages delivered to a promise are forwarded to their eventual fulfillment
value or rejected with the promises's eventual rejection reason.

Although Remotables and Promises are both Capabilities for delivering messages,
Promises do not support pass-invariant identity comparison.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,326 @@
...
+>   }
+>   ```
+> - **Python**: `Tagged(tag, value)` where `tag` is a `string` and `Tagged` is
+>   imported from `ocapn`.
+>
+> OCapN tagged data have a single value (are not variadic) [November 14, 2023](https://github.com/ocapn/ocapn/issues/94)
+>
+> What are tagged values for? https://github.com/ocapn/ocapn/issues/52
+
+## Remotable
```
kriskowal, line 263: @tsyesika @cwebber noted at the October 8, 2024 meeting that we should make this term consistent with the spec, either with spec changes or a change right here. Dave Thompson suggests the corresponding term throughout the spec at time of writing is “remote reference”. Agoric makes a distinction between a remote reference and a “remotable”: a remotable is a settled capability that _may_ be local or remote, not strictly remote.

I think that’s a distinction that applies to this abstract model, and it might be that we need to expand this section to indicate a distinction between a “local presence that forwards messages to a remote reference” and a “a local remotable object that may receive messages from local or remote objects”.

draft-specifications/Model.md
kriskowal, line 280:
```suggestion
> - **Guile**: a Goblins callable actor
```

draft-specifications/Model.md
kriskowal: Is this adequate?
```suggestion
> - **Guile**: a Goblins promise
```

draft-specifications/Model.md
kriskowal: How’s this?
```suggestion
> - **Guile**: an exception
```

draft-specifications/Model.md
kriskowal: @erights is proposing immutable ArrayBuffers this very week, so I anticipate updating this, but this is close enough for now.

draft-specifications/Model.md
davexunit: Maybe just "a Goblins actor"

draft-specifications/Model.md
```diff
@@ -0,0 +1,324 @@
...
+A value that can receive messages.
+
+A remotable value is either locally implemented or forwards messages to the peer
+where such a local implementation exists for delivery to it.
+The latter is referred to as a presence.
+
+Comparing the identities of remotable values is pass-invariant, meaning that
+comparing the identities of any pair of remotable values will produce the same
+result on one peer if passed over OCapN to another peer and compared there,
+regardless of whether either value is a presence on either peer.
```
tsyesika, line 273: As I'm reviewing this now, I have trouble wrapping my head around this sentence but that could just be the different terminology (compared to Goblins) and how these are implemented in JS. Not sure this is useful feedback but I thought I'd flag it anyway.

draft-specifications/Model.md
cwebber: The OCapN spec currently simply says "Remote References" which I prefer over "Remoteable" personally.

draft-specifications/Model.md
```diff
@@ -0,0 +1,324 @@
...
+> - **JavaScript** and **JSON**: `[]`
+> - **Python**: `()` (We have not discussed whether to use tuple or list. Tuple
+>   is less mutable, though enforcing immutability in Python is not likely to be
+>   a goal the way it is in hardened JavaScript.)
+>
+> Lists participate in the JSON subset of OCapN.
+>
+> We achieved consensus to name the type "List" at the [November 14, 2023](https://github.com/ocapn/ocapn/issues/94) meeting.
+
+## Struct (JSON)
```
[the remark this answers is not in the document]
kriskowal: The important point is that the keys must be strings, and “dictionary” does capture that distinction in juxtaposition with “map”. Agoric uses the term “record”, but I believe that was disqualified on the grounds that it is too close to a term present in the Syrup concrete model (but absent in the data model!). I am sure Agoric (@erights) would welcome the term “record” if we can move the Syrup word out of the way (say “tuple”).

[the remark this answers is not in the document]
kriskowal: If you ask a C person, they will tell you that a struct is an association between a name (constrained to valid C identifiers) and a value (at an offset in a compact memory representation implied by their order of appearance and padding necessary to satisfy the underlying architecture). Our interpretation is consistent with the germane portion of that.

If you ask a C++ person, a struct is simply a class where all of the fields and methods are implicitly public. As in most cases, the C++ person is wrong.

A Protobuf person will tell you they prefer the term “message”. Messages can contain messages, and in fact, you can’t express the difference between presence of a zero value and absence of a field without a so-called “message”. Their opinion can be dismissed because they had an opportunity to pick one of “protocol” or “interface”, and instead did both (Protobuf IDL) and added  “service”, to refer to essentially the same thing. Protobuf people have strange preferences.

If we choose to reöpen this aging wound, I dislike but am open to “dictionary”. I like “record”. I think “struct” is fine.

draft-specifications/Model.md
kriskowal: At the November 2024 meeting, we discussed this and resolved that “references” are a category divided between “promise” and “target”, where “targets” have a pass-invariant identity.

We came to this name because the root word “remote” implied that the important distinction between these references and promises is that they can be either “local” or “remote” (or “exported” and “imported” respectively), but that attribute does not distinguish them from promises. The term “target” is intended to capture both of the important distinctions:
- when delivering a message to a reference, the referent may be a promise, but the promise is not the target: it is a conveyance toward the target.
- regardless of whether the target is local or remote, the target has a unique and comparable identity, such that we can establish over OCapN that any pair of references ultimately refer to the same target. We sometimes call this “pass invariant” property of targets “grant matching”. Promises lack this property. In the physical metaphor of “target” a set of concentric circles imply a unique point in space toward which messages (on arrows) converge.

draft-specifications/Model.md
```diff
@@ -0,0 +1,324 @@
...
+> OCapN uses the name selector to avoid the implication that they will
+> correspond to a language's symbol type in all languages that have a symbol
+> type.
+> Selectors may correspond to symbols in languages where a symbol is eligible
+> for garbage collection when there are no extant references.
+> At this time, JavaScript cannot safely use registered symbols like
+> `Symbol.for('name')` for OCapN selectors, because registered symbols are
+> interned without possibility of eventual garbage collection.
+>
+> OCapN supports one operator for delivering both function application and
```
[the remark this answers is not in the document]
kriskowal: Extending the non-normative note in case the implication gets lost in the above about method invocation.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: Having discussed this here and with @erights out-of-band, I am continuing this PR with “struct” since we have tentative consensus on this and not the alternatives. Mark and I object to “dictionary” since our model for dictionaries in JavaScript/TypeScript is that a dictionary is a specialization on maps, where the keys are specifically strings, and values are a homogeneous type, whereas structs are modeled as objects with known and heterogeneous (KeyT to ValueT) pairs.

```ts
type Dict<T> = Map<string, T>
```

Mark and I would not object to “record” but expect Spritely would on grounds that the term is reserved for a different concept in the concrete Syrup protocol.

draft-specifications/Model.md
kriskowal: There are caveats.

- JSON documents that contain any string that includes a Unicode surrogate is not passable.
- JSON documents with numbers that exceed the precision available to an IEEE 64 bit floating point number are not passable.

The second caveat isn’t really a caveat for the ECMA 404 version of JSON but the Unicode limitation does not admit all JSON expressible in JavaScript.

This comment speaks more to the point that if OCapN didn’t have separate Null and Undefined, JSON documents that had an explicit `null` in them would not have pass-invariant structural equivalence (be “passable”). Null and undefined would have to be collapsed into a single type, such that either struct fields with null values were omitted when they returned, or struct fields will undefined values would return as null.

draft-specifications/Model.md
kriskowal: Are all Guile procedures eligible targets for message delivery or do they need to be marked with a Guile thing?

draft-specifications/Model.md
kriskowal: I’ve posted a commit that avoids the jargon “passable” and makes painfully clear the scenario that OCapN holds invariant by distinguishing Null from Undefined.

kriskowal:
I believe this to be ready for another round of review. All changes I’ve applied are in a second commit that I intend to be squashed before merging. In summary, I’ve responded to all above feedback as noted in replies, but particularly:

1. Added a summary of all value types, dividing them into 3 instead of 2 categories: Atom, Container, Reference, such that the two reference types can be clearly distinguished.
2. Added an appendix on the meaning of Pass Invariant Equality to unravel some unclear prose
3. Provided a more clear explanation of the invariant preserved with separate Null and Undefined. It is _narrow_.
4. I’ve used Reference instead of Capability throughout
5. I’ve used Target instead of Remotable throughout

draft-specifications/Model.md
```diff
@@ -179,6 +218,10 @@ Selectors cannot contain lone/unpaired surrogates.
 >   ```
 > - **Python**: `Selector('name')` where `Selector` is imported from `ocapn`.
 >
+> A selector's content must be expressible in UTF-8.
+> Strings in UTF-16 can express unpaired/lone surrogates that cannot be
+> expressed in any UTF and so cannot be carried by OCapN.
+>
 > Tracking: https://github.com/ocapn/ocapn/issues/46
 >
 > OCapN uses the name selector to avoid the implication that they will
```
erights, line 227:
```suggestion
> OCapN uses the name `selector` to avoid the implication that they will
```

draft-specifications/Model.md
```diff
@@ -235,50 +284,69 @@ A struct with unique, unordered string keys and values of heterogeneous type.
...
 > OCapN provides a small number of container types with minimal semantics.
 > Tags allow protocols to emerge from values annotated to indicate a richer
 > variety of semantics.
 > Concretely, a list might be differentiated from or hinted to be an ordered
 > list, sorted list, or a set of unique values.
 >
-> - **Guile**: currently a Guile record labeled `'(syrec ...)` (**imported**)
+> - **Guile**: currently a Guile record labeled `'(make-tagged 'tagName 10)` (**imported**)
 > - **JavaScript**: an object with the key `Symbol.for('passStyle')` and value
 >   `tagged`, with a string tagName and a single `"payload"` field carrying the tagged value.
```
erights, line 299: Good that the code example here and in `Selector` are similar. The text decorating that code should be similar too.

draft-specifications/Model.md
```diff
@@ -235,50 +284,69 @@ A struct with unique, unordered string keys and values of heterogeneous type.
...
+> We have not yet converged on any particular details about the modeling of
+> errors. The purpose of errors is typically to indicate that some requested
+> operation failed. The purpose of the contents of errors is to preserve and
+> convey diagnostic information, mostly to help debug problems, such as the
+> root cause of a surprising failure. This is a best-efforts obligation, for
+> which we have not yet decided either what contents are required, nor what is
+> allowed, nor what must be preserved as errors are passed from one site to
+> another. Until these details are decided, the only hard requirement is that
+> an error round trip to an error. We avoid any interpretation for now as to
+> whether it is the "same" error.
```
erights, line 331: Well said. Should have a tracking issue

draft-specifications/Model.md
```diff
@@ -235,50 +284,69 @@ A struct with unique, unordered string keys and values of heterogeneous type.
...
 
-A value that can receive messages.
+# Reference
```
erights, line 333: Saying `Reference` rather than `Capability` seemed good to me at first. But we name the protocol `ocapn` and make a big deal of it being a distributed capability protocol. This is where those capabilities appear. Calling these capabilities means there's one less concept to explain.

draft-specifications/Model.md
```diff
@@ -235,50 +284,69 @@ A struct with unique, unordered string keys and values of heterogeneous type.
...
 
-Remotable values can be either a local value or a remote value that forwards
-messages to a local value.
-The latter is a presence: a remotable that forwards messages to the peer where the
-message can be delivered to its corresponding local remotable.
+A value that can receive messages, either a Target or Promise.
```
erights, line 335: `Reference` and `Container` have an introductory one liner before the subheadings. Could we have one for `Atom` too?

draft-specifications/Model.md
```diff
@@ -292,6 +360,18 @@ Promises do not support pass-invariant identity comparison.
...
+A target might be sent from a local peer to a remote peer, then the remote peer
+may send that target back to the local peer.
+The sent target will be equal to the received target.
+
+> The operator for establishing equality is specific to the implementation
+> language.
+> For JavaScript, the Pass-Invariant Equal operator is `Object.is`.
+
+Although Targets and Promises are both References for delivering messages,
+OCapn does not maintain pass invariance for promises.
```
erights, line 373: There is no pass-invariant promise *identity* or *equality*. However you want to put it.

As with errors, promises must round trip only to promises. This is the testable part.

The less testable part, but still part of the pass invariance of the semantics:

A promise's eventual settlement must be pass invariant. IOW, if pa in peer a, when passed to peer b results in pb, then (possibly aside from disconnection/partition trauma), pa must eventually settle iff pb eventually settles. And their settlements must be the "same" up to pass invariance.

If we allow differential disconnection/partition trauma to result in promise rejection, then we'd instead make a similar statement about fulfillment:

A promise's fulfillment settlement must be pass invariant. IOW, if pa in peer a, when passed to peer b results in pb, then, if either pa or pb eventually fulfills, the other must either eventualy fulfill to the "same" value up to pass invariance, or it must reject (with a reason indicating disconnection/partition? Do we want ocapn to stdize such a distinction?)

draft-specifications/Model.md
```diff
@@ -235,50 +284,69 @@ A struct with unique, unordered string keys and values of heterogeneous type.
...
 
-> - **Guile**: a function
-> - **JavaScript**: A *far object* or a *remote presence* of that far object.
+> - **Guile**: a procedure
+> - **JavaScript**: Tentatively, *far object* (can handle message deliveries
+>   locally) or a *presence* (a token object that forwards messages to a
+>   corresponding far object).
 >   ```js
 >   {
 >     __proto__: {
```
erights, line 352: Need to say something about the `__proto__`. Also, the actual remotable object is an object that inherits from the object shown, possibly through multiple levels, each contributing their own methods.

Again, would be good for all such code examples to have similar enough text to help people reason by similarity and difference.

draft-specifications/Model.md
```diff
@@ -306,21 +386,26 @@ value or rejected with the promises's eventual rejection reason.
 >
 > Tracking https://github.com/ocapn/ocapn/issues/55
 
-## Error
+OCapN does not maintain [Pass Invariant Equality](#pass-invariant-equality) for
+promises.
```
erights, line 390: Yes, this says it correctly.

draft-specifications/Model.md
```diff
@@ -306,21 +386,26 @@ value or rejected with the promises's eventual rejection reason.
...
+
+## Pass Invariant Equality
+
+A type holds Equality invariant over passage between OCapN peers if values of that
+type can be passed from a local peer to a remote peer, then returned to the local peer,
+and the value that was sent is Equal to the value received.
+
+> The operators necessary for checking Equality may vary for each language and even
+> for types in one language.
+> For example, `Object.is`
```
erights, line 410: Actually, you should use `Object.is` as a clarifying counter-example. Containers do ***NOT*** have pass invariant `Object.is` equality.

erights (changes requested):

draft-specifications/Model.md
```diff
@@ -0,0 +1,411 @@
...
+> type.
+> Selectors may correspond to symbols in languages where a symbol is elligible
+> for garbage collection when there are no extant references.
+> At this time, JavaScript cannot safely use registered symbols like
+> `Symbol.for('name')` for OCapN selectors, because registered symbols are
+> interned without possibility of eventual garbage collection.
+>
+> OCapN supports one operator for delivering both function application and
+> method invocation.
+> By convention, method invocation is equivalent to function application, where
```
zarutian, line 238: is the method selector (first argument) omitted in the case of function invocation?

in Spritely the method selector (first argument) is #f in function invocation, in E __run was used and in Monte it was run.

Thoughts?

erights: > in Spritely the method selector (first argument) is #f in function invocation

If that's true, then we have badly misunderstood how these concepts map to spritely, to a degree that does likely invalidate what we thought we agreed to.

@cwebber , @tsyesika , please clarify. Thanks.

tsyesika: It’s not true. When invoking it as a function you simply provide the function arguments as the arguments. No need to pass false or do anything else.

kriskowal: Pardon, this is stale. I’ll update.

zarutian: I thought it was so on the 'wire' of a Spritely captp connection.

draft-specifications/Model.md
kriskowal: I began the adventure of writing in pass-invariant equality for every model type and quickly mired in details and exceptions that are outside consensus. We can’t yet say that Errors (which are Containers) are not pass-invariant by `Object.is` (lacking consensus about errors, which either need to be or be associated with a capability) and we also can’t say that ByteArrays (Atoms) are pass-invariant by `Object.is`. I believe we need a new issue on Pass-Invariant Equality and to fold that detail in follow-up. To that end, I can leave Pass-Invariant Equality for References or pull it out for a later holistic treatment.

draft-specifications/Model.md
kriskowal: We could also add a case to `pass-style` such that any `x` where `typeof x === 'function'` is a valid target and revise `Far` and `exo` to use that as their foundation, obviating this detail.

draft-specifications/Model.md
```diff
@@ -109,20 +144,25 @@ OCapN provides no support for other floating point precisions.
...
 
 > - **Guile**: `""`
 > - **JavaScript**: `''`
 > - **Python**: `''`
 >
+> A string's content must be expressible in UTF-8.
+> Strings in UTF-16 can express unpaired/lone surrogates that cannot be
+> expressed in any UTF and so cannot be carried by OCapN.
+>
 > \*Strings participate in the JSON subset of OCapN except any strings that
```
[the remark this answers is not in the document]
kriskowal: Some flavors of Markdown get confused about whether a literal `*` indicates a bullet. I’ll change these to daggers† to avoid the problem.

draft-specifications/Model.md
erights: I do not believe we should do that, but it is arguable. I'll  leave the reasons to a verbal discussion sometime.

erights: In any case, it would be a huge change that we're not likely to consider anytime soon. It would be good to have the JS examples reflect current endo reality.

draft-specifications/Model.md
tsyesika: > I thought it was so on the 'wire' of a Spritely captp connection.

There used to be a dedicated method field on desc:deliver which was set to false on function invocation but it was never really used iirc and was dropped a long time ago

draft-specifications/Model.md
erights: If we cannot settle this here, then we should postpone to a later PR. But I believe we've already committed to clear answers to all of these, with the possible exception of errors, and I'm happy with those answers:

- The pass-invariant ocapn equality of ***Targets*** is correctly reflected in JS as `Object.is`

- The pass-invariant ocapn equality of ***Atoms*** that are represented in JS by ***JS primitives*** is correctly reflected in JS as `Object.is`.

- The pass-invariant ocapn equality of ***ByteArrays***, because they are represented in JS as non-JS primitives that we are *not* going to intern, cannot be reflected in JS as `Object.is`. Note that this awkward difference in the JS behavior associated with ocapn Strings vs ocapn ByteArrays affects only the JS programmer. No such asymmetry bleeds into the ocapn level.

- The pass-invariant ocapn equality of ***Containers*** --- Lists, Structs, Taggeds --- is most definitely not reflected in JS as `Object.is`.

- OCapN ***Promises*** have no pass-invariant ocapn equality, so the question of `Object.js` applied to the JS representation of an ocapn promises does not arise. Other pass-invariant aspects of promises, like that they remain promises, and whatever we specify wrt observable settlement, are not *equality* invariants.

IMO, the only hard one:

- OCapN ***Errors*** have no pass-invariant ocapn equality, so the question of `Object.js` applied to the JS representation of an ocapn errors does not arise. Other pass-invariant aspects of errors, like that they remain errors, and that at least some of their properties (like `name` but TBD) are considered semantic and must round trip, are not *equality* invariants.

The reason the Error case is so problematic is that, for diagnostic purposes only, they carry an associated identity (variously referred to an an `errorId`) for looking up associated out-of-band diagnostic info. The naming of this identity will somehow need to be reflected in the ocapn wire protocol. It remains to be seen whether we succeed at specifying this while maintaining a clear separation between semantics and diagnostics. If this separation collapses, errors may indeed come to have cap-like unforgeable identity. But I really hope we can keep this distinction, and IMO we should proceed to try because we may succeed.

But, because of the remaining uncertainty about errors, perhaps this part alone should be postponed to a later PR.

erights: Having written out the above response, I'm wondering about this new classification of Errors as Containers as well. For now, until we understand errors better, what do you think of just having them be their own distinct category outside of Atom, Container, Reference/Capability?

draft-specifications/Model.md
erights: Simple (and IMO adequate) reason not to consider unmarked functions to be targets by default:

We are trying really hard to shift 
- from a local E-like hardened js programming model where every object and function boundary should be a defensive boundary
- to a local endo programming model where, in general, the defensive boundaries are target invocation boundaries, where targets are typically exos protected by interface guards.

For example, one of the reasons I find our shift towards more direct support of ts acceptable, despite typescript's unsound type system, is that to the exo programmer, the relevant type-like system for defensive exo boundaries will be the declarative interface guards, where the ts types directly inferred from these guards are both static and sound.

Admitting unmarked functions by default as targets would completely undo the progress we've made towards this coarsening of defensive boundaries.

draft-specifications/Model.md
zarutian: s/desc:deliver/op:deliver/ no?

draft-specifications/Model.md
```diff
@@ -109,20 +144,25 @@ OCapN provides no support for other floating point precisions.
...
-A string of Unicode code points but no lone/unpaired surrogates.
+A string of Unicode code points excluding surrogates (U+D800-U+DFFF).
+Strings are distinguished from [Selector](#selector) s by type, not content.
 
 > - **Guile**: `""`
 > - **JavaScript**: `''`
 > - **Python**: `''`
 >
+> A string's content must be expressible in UTF-8.
+> Strings in UTF-16 can express unpaired/lone surrogates that cannot be
```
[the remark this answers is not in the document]
kriskowal: I’m cleaning this up to just say “some two-byte representations including JavaScript strings”.

draft-specifications/Model.md
kriskowal: Chose to TitleCase as that’s the prevailing style.

draft-specifications/Model.md
kriskowal: Adding a link to https://github.com/ocapn/ocapn/issues/142

draft-specifications/Model.md
kriskowal: Pending feedback from @cwebber and @tsyesika, I’ll for now use Reference as the primary name and Capability as its epithet in parentheses.

draft-specifications/Model.md
kriskowal: If you will allow it to be circular 😉.

draft-specifications/Model.md
kriskowal: I’ll retract the proposition that we allow all bare JavaScript functions to be valid targets.

A problem with documenting Endo reality here is that it is not sufficient to express a target that can receive an arbitrary delivery. We could document Endo reality as a special case, but the general case ought to be a marked function, e.g.,

```js
({
  [Symbol.for('passStyle')]: 'target',
  target: (...args) => {},
})
```

Given that these non-normative sections are for examples and not per-language specifications, I find that the general case is sufficient for now, provided that you concur.

draft-specifications/Model.md
kriskowal: I’ll tentatively move Error into its own category, pending consensus on whether it is a container or reference.

draft-specifications/Model.md
kriskowal: Okay, now seated and looking at what I’ve written, the proposed specification does not suggest that non-method invocation must have False as the first argument. The text is consistent with consensus. I do recall the former idea about `#f`, but it is not current.

kriskowal:
I’ve posted a subsequent commit that clarifies the various pass invariants and makes them explicit for every model type. I’ve also factored out the JSON invariants and referenced them for the two applicable caveats for Null and Float64. I also used the tentative aspirational example for marked targets in JavaScript, such that they can handle all deliveries, although this is not consistent with Endo’s implementation yet. For @erights consideration in particular.

draft-specifications/Model.md
kriskowal: I’ve posted a commit that regularizes the JavaScript examples in the cases where they’re objects with a passStyle.

draft-specifications/Model.md
```diff
@@ -22,22 +49,30 @@ A value representing the absence of a value.
 
 A value representing `null` as distinct from `undefined` in JSON.
```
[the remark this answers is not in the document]
kriskowal: I’ve made this more specific and linked a section on JSON invariants.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+
+- Any JSON document that a JavaScript peer can produce using
+  `JSON.stringify`,
+- excluding those that contain any Unicode surrogates,
+- can then be read back with `JSON.parse`,
+- sent to any other OCapN peer,
+- returned from any OCapN peer,
+- then saved back to an identical JSON document with `JSON.stringify`.
+
+> For this reason, OCapN supports distinct a distinct [Undefined](#undefined)
```
erights, line 589:
```suggestion
> For this reason, OCapN supports a distinct [Undefined](#undefined)
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+OCapN does not maintain [Pass Invariant Identity](#pass-invariant-identity) nor
+[Pass Invariant Equality](#pass-invariant-equality) for promises.
+
+As with all other types, OCapN does maintain [Pass Invariant
+Type](#pass-invariant-type) such that a local promise sent to a a remote peer
+arrives as a promise, regardless of whether it is pending or settled locally.
+
+OCapN maintains that, if a local promise is sent to a remote peer and then
+retuned, either the fulfillment value or the rejection reason of the sent and
+received promises will satisfy the pass invariants applicable to their type.
```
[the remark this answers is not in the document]
erights: Unless "and then returned" serves a purpose I'm missing, which is quite possible.

draft-specifications/Model.md
gibson042: > * JSON documents that contain any string that includes a Unicode surrogate is not passable.
> * JSON documents with numbers that exceed the precision available to an IEEE 64 bit floating point number are not passable.

Annals of "everything is exactly the same": [RFC 7493 I-JSON](https://www.rfc-editor.org/rfc/rfc7493)
> Object member names, and string values in arrays and object members, MUST NOT include code points that identify Surrogates or Noncharacters as defined by [<a href="https://www.rfc-editor.org/rfc/rfc7493#ref-UNICODE" title="&quot;The Unicode Standard&quot;">UNICODE</a>].
> …
> I-JSON messages SHOULD NOT include numbers that express greater magnitude or precision than an IEEE 754 double precision number provides, for example, 1E400 or 3.141592653589793238462643383279.
> …
> An I-JSON sender cannot expect a receiver to treat an integer whose absolute value is greater than 9007199254740991 (i.e., that is outside the range [-(2\*\*53)+1, (2\*\*53)-1]) as an exact value.
> …
> Objects in I-JSON messages MUST NOT have members with duplicate names.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+>
+> Tracking: https://github.com/ocapn/ocapn/issues/49
+
+Targets have [Pass Invariant Identity](#pass-invariant-identity).
+A target might be sent from a local peer to a remote peer, then the remote peer
+may send that target back to the local peer.
+The sent target will be identical to the received target and no other value.
+
+OCapN does not impose [Pass Invariant Equality](#pass-invariant-equality) on
+any [Reference](#reference).
```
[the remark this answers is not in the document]
gibson042:
```suggestion
OCapN does not impose [Pass Invariant Equality](#pass-invariant-equality) on
any [Reference](#reference-capability).
```
(but independent of that, I agree with @erights—I think References _do_ require Pass Invariant Equality, and AFAICT Pass Invariant Identity should be defined to logically imply Pass Invariant Equality).

gibson042 (commented):
Mostly editorial suggestions, including phrasing, typo fixes, and Markdown rendering improvements. Note also the general pattern that generated heading ids include all alphabetic text content (e.g., `# Reference (Capability)` corresponds with slug "#reference-capability" rather than "#reference").

draft-specifications/Model.md
erights: What?

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+
+## Promise
+
+A promise represents the eventual return value (fulfillment) or thrown error
+(rejection reason) for a message delivery.
+A promise is pending until settled with either a fulfillment or rejection reson.
+Messages delivered to a promise are forwarded to their eventual fulfillment
+value or rejected with the promises's eventual rejection reason.
```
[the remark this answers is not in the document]
erights: Noting that we do not do this now, at least not always, and might never choose to do so. Though I agree that the meaning of deliverying a message to a non-ocapn-target is outside the meaning of ocapn. For the endo and swingset ocapn language binding implementation, atoms and containers might still have js method behaviors, and I think our systems will, at least sometimes, invoke these when a promise fulfills to one.

Would be good to test.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+> Target, the sent and received targets will be identical, because Targets
+> maintain [Pass Invariant Identity](#pass-invariant-identity).
+>
+> For another example, if the sent promise settles with a fulfillment value
+> that is a Struct, the sent and received structs will be equal, because
+> all [Containers](#container) maintain [Pass Invariant
+> Equality](#pass-invariant-equality), but the structs may not be identical.
+> OCapN doesn't explicitly forbid identity for passed structs, but interning
+> structs for the duration of an OCapN session in order to preserve identity
+> for all equivalent structs is prohibitively expensive in most languages.
```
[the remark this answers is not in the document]
erights: Still seems consistent with the idea that "identity" is not usefully an ocapn concept, but rather useful to only describe expected features of language bindings.

erights (changes requested):
Everything else still looks good enough to me, but I went back to "Request changes" because of the aspirational representation choice shown for a JS target, which I doubt is close to anything we'll actually want to do. Suggest discussing this at Endo meeting on 11 December. But until then please show the representation we actually use.

draft-specifications/Model.md
tsyesika: So I guess there's a bit of a blurring of the lines between Guile and Goblins. In terms of references, Goblins has two object/actor references:

`<local-object-refr>` this is a reference which refers to a guile procedure (the object's/actor's behaviour) that was returned by spawning the actor's constructor.
`<remote-object-refr>` which is a reference to an actor on a remote system.

In addition to those, you can also send messages to promises, there are two equivalent promise references:
- `<local-promise-refr>`
- `<remote-promise-refr>`

When Goblin's CapTP layer gets a message to a `<local-object-refr>` it'll queue that message on the vat's message queue so that the vat invokes that procedure in turn.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+>   for the string name `deliver`, the value is a function that receives the
+>   delivery arguments and returns or throws the resolution of the delivery's
+>   promise.
+>   ```js
+>   ({
+>     [Symbol.for('passStyle')]: 'target',
+>     [Symbol.toStringTag]: allegedInterfaceName,
+>     deliver(...args) {},
+>   })
+>   ```
```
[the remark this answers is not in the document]
kriskowal: In preparation for seeking adoption of this model at the December 10 OCapN meeting, I am tentatively removing the JavaScript example ("to be proposed"). Reminder: all examples are non-normative comments.

draft-specifications/Model.md
```diff
@@ -0,0 +1,324 @@
...
+> - **Python**: `''`
+>
+> \*Strings participate in the JSON subset of OCapN except any strings that
+> contain unpaired/lone surrogates.
+>
+> OCapN cannot send or receive strings that come from a UTF-16 language
+> representation (such as a JavaScript string) that contains lone/unpaired
+> surrogates.
+> Notes: [Jan 2024 meeting
+> notes](https://cryptpad.fr/code/#/2/code/edit/Ng7kXl9rCF5qnTQX60UCpDHE/)
```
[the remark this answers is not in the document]
turadg: fwiw, that link now says "This document was deleted because of inactivity"

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: Thanks. I was confused. I am merging Identity and Equality into Equality. They did not overlap.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: I’m revising this to be more specific. Promises forward messages only to the fulfillment value if that value is a Reference/Capability. I did not intentionally imply messages would be forwarded to values and containers, which would leak language-specific protocols.

kriskowal:
I’ve adopted editorial recommendations, merged Identity and Equality into Equality, and removed the example for JavaScript targets ('to be proposed') since the current implementation of Endo CapTP does not reflect an ability to receive arbitrary deliveries and so doesn’t make a good example. When we have a way for a target to handle arbitrary deliveries (where the first argument is not a selector), I intend to update. That will also provide a suitable basis for the `echo` function we will need to test round-tripping of all deliverable messages.

draft-specifications/Model.md
```diff
@@ -395,9 +368,6 @@ Tagged values are equal or equivalent for the purposes of [Pass Invariant
 Equality](#pass-invariant-equality) if the tag and value are equivalent,
 transitively.
```
erights, line 369: Well, now that we've gotten rid of the unnecessary distinction between identity and equality, can we also dispense with "equivalent" and just speak of "equality"? What does talk of equivalence add to the meaning of this sentence?

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
 
-A type holds equality invariant over passage between OCapN peers if values of that
-type can be passed from a local peer to a remote peer, then returned to the local peer,
-and the value that was sent is equal to the value received.
+OCapN defines a flavor of equality, nominally Equality.
+Two values may be Equal under Equality.
+Values of different OCapN types are not Equal.
+Values of the same OCapN type may be Equal under conditions specific to their
+shared type.
+All OCapN types maintain pass invariant Equality except Promise.
```
erights, line 480:
```suggestion
All OCapN types maintain pass invariant Equality except Promise and (at least for now) Error.
```

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
 
-> Equality does not guarantee a unique identity for all equivalent values.
+All OCapN implementations, regardless of language, must model all OCapN types
+such that Equality for the in-language representation of values is consistent
+with the Equality and Pass Invariance of Equality for their corresponding OCapN
+model type.
```
erights, line 485: Why? This seems both wrong as a concrete requirement and none of ocapn's business. It should be up to each language binding.

It is certainly wrong as stated, because it seems to refer to one of the language's existing notions of equality. For the endo binding, there is no JavaScript-level equality concept that corresponds to ocapn equality. Rather, the endo ocapn binding itself provides new equality predicates that, together, do correspond to ocapn equality. However, at least currently, we do not provide a single equality predicate that does so directly.

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
 
-Any pair of values of different types are not equal or equivalent regardless of
-their values.
+A type holds Equality invariant over passage between OCapN peers because any
+pair of local values that are Equal can be passed to another and the remote
+values will also be Equal.
```
erights, line 489: I'm not sure what this sentence is trying to mean.

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
 
-> Only values of the same type can be equivalent, per the equivalence relation
-> specified for their common type.
+It follows that, for any local value that is sent to a remote peer and then returned,
+the sent and received values will be Equal.
 
-OCapN holds equality invariant for all [Atoms](#atom) and
-[Containers](#container) passed between peers.
+> OCapN Equal values may be locally distinguishable by oother operators
```
erights, line 494:
```suggestion
> OCapN Equal values may be locally distinguishable by other operators
```

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
-OCapN does not hold identity invariant for values of type [Promise](#promise).
-
-OCapN does not guarantee identity is invariant for any values of [Atom](#atom)
-or [Container](#container) types.
+> distinction between 0 and -0 and lack of distinction between any two NaN
+> values.
+> `Object.is` is also sufficient for Target provided that the OCapN
+> implementation holds invariant a single object identity for every Target.
+> The JavaScript ArrayBuffer representation of an OCapN ByteArray must be
+> compared byte-for-byte.
```
erights, line 506: Good! 

Also, `Object.is` doesn't work on containers because container equality is based on recursive equality of the contents.

draft-specifications/Model.md
```diff
@@ -526,69 +472,56 @@ local peer, the sent and received values will have the same type.
...
+- Any JSON text that a JavaScript peer can produce using
   `JSON.stringify`,
-- excluding those that contain any Unicode surrogates,
+- excluding those in which the Unicode representation of any object member name
+  or string includes a surrogate,
 - can then be read back with `JSON.parse`,
 - sent to any other OCapN peer,
 - returned from any OCapN peer,
-- then saved back to an identical JSON document with `JSON.stringify`.
+- then saved back to an identical JSON text with `JSON.stringify`.
```
erights, line 520:
```suggestion
- then saved back to an equivalent JSON text with `JSON.stringify`.
```

Even though `JSON.stringify` is much more deterministic than the JSON standard demands, it is not canonical. It much encode the field of the same struct in a different order.

draft-specifications/Model.md
```diff
@@ -179,10 +170,9 @@ For purposes of [Pass Invariant Equality](#pass-invariant-equality):
...
 > `==` and `===` are not.
 
-OCapN does not impose [Pass Invariant Identity](#pass-invariant-identity) on
-any [Atom](#atom).
+## String
 
-## String ([JSON](#json-invariants)†)
+([JSON](#json-invariants)†)
 
 A string of Unicode code points excluding surrogates (U+D800-U+DFFF).
```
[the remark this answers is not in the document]
gibson042: This is correct as written, but it's true that people often confuse code points and code units. Perhaps a note?
> Well-formed UTF-16 can include 16-bit code _units_ in the inclusive range 0xD800-0xDFFF, but only as part of a surrogate pair which encodes a single code point greater than or equal to U+10000. Any code unit sequence including an unpaired surrogate is ill-formed and thus not an OCapN String (i.e., each OCapN string can be encoded into a well-formed Unicode string in any encoding form).

draft-specifications/Model.md
```diff
@@ -357,9 +333,6 @@ transitively.
 > OCapN implementations will likely sort keys of structs according to code
 > points in order in the concrete representation of a struct.
```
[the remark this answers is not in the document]
kriskowal: Removing.

draft-specifications/Model.md
kriskowal: I’m just going to remove for now, for tentative lack of verbiage that expresses my intent. I’m trying to express a constraint about a transitive round-trip invariant.

draft-specifications/Model.md
kriskowal: Riffing on this:
```
A type holds Equality invariant over passage between OCapN peers.
Any pair of local values that are Equal can be passed to another peer and the
respective remote values will also be Equal.

It follows that, for any local value that is sent to a remote peer and then returned,
the sent and received values will be Equal.
```

draft-specifications/Model.md
kriskowal: I do not regret this keyboard layout, because it ameliorates my emacs-claw, but it does have a cost.

draft-specifications/Model.md
```diff
@@ -0,0 +1,546 @@
...
+> order, though this is not a concern of the abstract syntax and data model.
+>
+> A JavaScript object that owns any non-string keys (numbers or symbols) or
+> with a prototype other than `Object.prototype` cannot be passed as an OCapN
+> struct.
+> The key `Symbol.for('passStyle')` is special and indicates the kind of OCapN
+> value the object represents.
+
+A pair of structs are equal or equivalent for purposes of [Pass Invariant
+Equality](#pass-invariant-equality) if they mutually posess a value for every
```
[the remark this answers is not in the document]
kriskowal: Insert _posse comitatus_ pun.

draft-specifications/Model.md
kriskowal: Added.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: I have gone back and forth on this. “Excluding surrogates” is a stronger normative requirement, albeit less clear.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+
+# Value
+
+A value is any [Atom](#atom), [Container](#container), [Reference](#reference),
+or [Error](#error).
+
+Atoms:
+
+- [Undefined](#undefined)
+- [Null](#null)
```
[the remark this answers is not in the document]
kriskowal: I moved the reference into the first line of the body of the section to uncomplicate the links.

draft-specifications/Model.md
erights: I suspected that was what you were trying to get at. How about we talk about a language binding as a pair of functions, (naming? marshal/unmarshal, encode/decode, serialize/unserialize, f/g ?) where 
- f(passable-lanuage-value) -> ocapn-value
- g(ocapn-value) -> passable-language-value

and then state the round tripping requirement for the language side as 

for all ocapn-values `ov`, `ocapn-equal(ov, f(g(ov)))`

IOW, they round trip through the language-binding preserving ocapn equality.

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+> JavaScript's `Object.is` is consistent with Pass Invariant Equality, whereas
+> `==` and `===` are not.
+
+## String
+
+([JSON](#json-invariants)†)
+
+A string of Unicode code points excluding lone (or unpaired) surrogates
+(U+D800-U+DFFF).
+Strings are distinguished from [Selector](#selector) s by type, not content.
```
gibson042, line 180: Surrogate code _points_ can never be "paired".
```suggestion
A sequence of Unicode code points excluding surrogates (U+D800-U+DFFF).
An implementation of OCapN in JavaScript or another language in which
strings are based on UTF-16 may include 16-bit surrogate code _units_
(0xD800-0xDFFF), but only as part of a surrogate pair which encodes a
single code point greater than or equal to U+10000 (i.e., each OCapN
string can be encoded into a well-formed Unicode string in any encoding
form).

Strings are distinguished from [Selector](#selector)s by type, not content.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+any [Atom](#atom).
+
+## String ([JSON](#json-invariants)†)
+
+A string of Unicode code points excluding surrogates (U+D800-U+DFFF).
+Strings are distinguished from [Selector](#selector) s by type, not content.
+
+> - **Guile**: `""`
+> - **JavaScript**: `''`
+> - **Python**: `''`
```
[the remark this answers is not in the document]
gibson042:
```suggestion
> - **Guile** and **JSON**: `""`
> - **JavaScript**: `''` or `""`
> - **Python**: `''` or `""`
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+A string of Unicode code points excluding lone (or unpaired) surrogates
+(U+D800-U+DFFF).
+Strings are distinguished from [Selector](#selector) s by type, not content.
+
+> - **Guile**: `""`
+> - **JavaScript**: `''`
+> - **Python**: `''`
+>
+> †Strings participate in the [JSON subset](#json-invariants) of OCapN except
+> any strings that contain unpaired/lone surrogates.
```
gibson042, line 187:
```suggestion
> †Strings participate in the [JSON subset](#json-invariants) of OCapN except
> any strings that contain surrogate code points.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+> predicate](https://github.com/tc39/proposal-is-usv-string), then it is not a
+> Passable string.
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), a pair of
+Strings are equal if they have the same quantity of Unicode code points and
+have the same respective Unicode code points in order.
+
+## ByteArray
+
+An array of bytes.
```
gibson042, line 208: This may be _too_ pedantic, but...
```suggestion
An array of 8-bit bytes.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+
+OCapN does not impose [Pass Invariant Identity](#pass-invariant-identity) on
+any [Atom](#atom).
+
+> Notably, imposing identity invariants for JavaScript ArrayBuffer would put
+> an undue performance burden for OCapN sessions, as they would need to intern
+> all reachable ByteArrays for the duration of the session and search for
+> an equivalent ArrayBuffer for every received ByteArray.
+
+## Selector
```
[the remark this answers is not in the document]
gibson042: Nit: Given the similarity of their definitions, I would find it convenient for Selector to appear adjacent to String, moving ByteArray out from in between them.

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), a pair of
+ByteArrays are equal if they have the same quantity of bytes and have the same
+respective bytes in order.
+
+## Selector
+
+A string of Unicode code points excluding lone (or unpaired) surrogates
+(U+D800-U+DFFF).
+Selectors are distinguished from [String](#string)s by type, not content.
```
gibson042, line 239:
```suggestion
A sequence of Unicode code points excluding surrogates (U+D800-U+DFFF).
An implementation of OCapN in JavaScript or another language in which
strings are based on UTF-16 may include 16-bit surrogate code _units_
(0xD800-0xDFFF), but only as part of a surrogate pair which encodes a
single code point greater than or equal to U+10000 (i.e., each OCapN
string can be encoded into a well-formed Unicode string in any encoding
form).

Selectors are distinguished from [String](#string)s by type, not content.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+## Selector
+
+A string of Unicode code points excluding surrogates (U+D800-U+DFFF).
+Selectors are distinguished from [String](#string)s by type, not content.
+
+> - **Guile**: symbols `'name`
+> - **JavaScript**: an object with two own properties:
+>   for the registered symbol key `passStyle`, the value is the string
+>   `selector`; and
+>   for the well-known symbol `toStringTag`, the value is the selector.
```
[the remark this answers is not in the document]
gibson042:
```suggestion
>   for the well-known symbol `toStringTag`, the value is a string containing
>   the code points of the selector.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+>   ```js
+>   ({
+>     [Symbol.for('passStyle'): 'selector',
+>     [Symbol.toStringTag]: 'name',
+>   })
```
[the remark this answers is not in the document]
gibson042:
```suggestion
>   ({
>     [Symbol.for('passStyle')]: 'selector',
>     [Symbol.toStringTag]: 'name',
>   })
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+>   ```
+> - **Python**: `Selector('name')` where `Selector` is imported from `ocapn`.
+>
+> A selector's content must be expressible in UTF-8.
+> Strings in UTF-16 can express unpaired/lone surrogates that cannot be
+> expressed in any UTF and so cannot be carried by OCapN.
```
gibson042, line 256: UTF-16 does not allow unpaired surrogates, but the arbitrary 16-bit UCS-2/"WTF-16" encoding used in e.g. JavaScript does.

```suggestion
> Some two-byte Unicode encodings, as in JavaScript strings, can contain
> unpaired/lone surrogates that have no valid expression in any UTF and so
> cannot be carried by OCapN.
```

draft-specifications/Model.md
gibson042:
```suggestion
> Selectors may correspond to symbols in languages where a symbol is eligible
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+> Tracking: https://github.com/ocapn/ocapn/issues/46
+>
+> OCapN uses the name Selector to avoid the implication that they will
+> correspond to a language's symbol type in all languages that have a symbol
+> type.
+> Selectors may correspond to symbols in languages where a symbol is elligible
+> for garbage collection when there are no extant references.
+> At this time, JavaScript cannot safely use registered symbols like
+> `Symbol.for('name')` for OCapN selectors, because registered symbols are
+> interned without possibility of eventual garbage collection.
```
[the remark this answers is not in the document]
gibson042:
```suggestion
> `Symbol.for('name')` for OCapN selectors, because some implementations intern
> such symbols without possibility of eventual garbage collection.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+A pair of lists are equal or equivalent for purposes of [Pass Invariant
+Equality](#pass-invariant-equality) if the are the same length and every
+respective value is equal, transitively.
+
+OCapN does not impose [Pass Invariant Identity](#pass-invariant-identity) on
+lists.
+
+## Struct ([JSON](#json-invariants))
+
+A struct with unique, unordered string keys and values of heterogeneous type.
```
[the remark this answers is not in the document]
gibson042:
```suggestion
A collection of unordered (key, value) pairs.
Each key must be a string, and must be unique within a Struct.
Values within a Struct may be of heterogeneous type.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+> points in order in the concrete representation of a struct.
+
+OCapN does not impose [Pass Invariant Identity](#pass-invariant-identity) on
+any [Container](#container).
+
+## Tagged
+
+A [Value](#value) with a tag.
+The value may be any value.
+The tag is a string of Unicode code points excluding surrogates (U+D800-U+DFFF).
```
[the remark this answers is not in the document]
gibson042:
```suggestion
A pair consisting of a tag and a [Value](#value).
The tag must be a [String](#string).
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+such that Equality for the in-language representation of values is consistent
+with the Equality and Pass Invariance of Equality for their corresponding OCapN
+model type.
+
+A type holds Equality invariant over passage between OCapN peers.
+Any pair of local values that are Equal can be passed to another peer and the
+respective remote values will also be Equal.
+
+It follows that, for any local value that is sent to a remote peer and then returned,
+the sent and received values will be Equal.
```
gibson042, line 494: https://github.com/ocapn/ocapn/pull/125#discussion_r1854557565
> I don't think this characterization is quite strong enough—wouldn't it be incorrectly satisfied if a remote peer were to factor in the source from which it receives a value? We want to cover invariance on _all_ peers.

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+> each value.
+> For example, Float64 can be compared by `Object.is`, which preserves the
+> distinction between 0 and -0 and lack of distinction between any two NaN
+> values.
+> `Object.is` is also sufficient for Target provided that the OCapN
+> implementation holds invariant a single object identity for every Target.
+> The JavaScript ArrayBuffer representation of an OCapN ByteArray must be
+> compared byte-for-byte.
+> Also, Object.is doesn't work on Container because container equality is based
+> on recursive equality of the contents.
```
gibson042, line 510:
```suggestion
> Also, `Object.is` is not sufficient for Container because container
> equality is based on recursive equality of the contents.
```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+
+# JSON Invariants
+
+OCapN holds invariant that:
+
+- Any JSON text that a JavaScript peer can produce using
+  `JSON.stringify`,
+- excluding those in which the Unicode representation of any object member name
+  or string includes a surrogate,
```
gibson042, line 519:
```suggestion
- excluding those in which the Unicode representation of any object member name
  or string includes a surrogate code point,
```

gibson042 (approved):

draft-specifications/Model.md
kriskowal: Endo/Agoric have analogous local and remote targets, though we just use a JavaScript promise for promises regardless of remote vs local. We call a remote target a “presence” which is just an opaque object that we can send messages to.

Follow-up question: Does Goblins oblige you to wrap a procedure to construct a `<local-object-refr>` in order for it to be able to receive deliveries? In JavaScript, we mark objects with `makeExo(interfaceName, interfaceGuard, methods)`, which is a user mode utility for generating an object that we recognize as an eligible target, as opposed to arbitrary JavaScript objects. This distinction prevents the JavaScript API from implicitly leaking into protocols above CapTP. I imagine this is less necessary for Goblins, but your answer may come to bear on the accuracy of this non-normative example.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: The cryptpad links have been replaced with github permalinks.

draft-specifications/Model.md
kriskowal: I’ve rewritten this section. Please let me know if it continue to be opaque.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: At the December 10 meeting, we decided to proceed with verbiage that the name “struct” is tentative. https://github.com/ocapn/ocapn/issues/144

draft-specifications/Model.md
kriskowal: I removed the back-ticks and promoted the case of “selector” to “Selector” to denote a pronoun, which I’m now using consistently to distinguish casual nouns from nouns with OCapN definitions.

draft-specifications/Model.md
kriskowal: For the record, @erights does not concur and I’ve removed the example, since no example that is sufficiently expressive to receive an arbitrary delivery yet exists for JavaScript.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: I’ve revisited this and broken it into corollaries. The first is maintaining the invariant over passage. The second implying invariance for a round-trip.

draft-specifications/Model.md
kriskowal: I’ve doctored this into the non-normative qualifier below the example and synced it with selector instead.

```
> A string/selector's content must be expressible in UTF-8.
> Some two-byte Unicode encodings, as in JavaScript strings, can contain
> 16-bit surrogate code _units_ in the range from 0xD800-0xDFFF.
> Pairs of surrogate code units correspond to a single Unicode code _point_
> greater than or equal to U-10000 and can be expressed in UTF-8.
> However unpaired or lone surrogates have no valid expression in any UTF
> and so cannot be carried by OCapN.

```

draft-specifications/Model.md
```diff
@@ -0,0 +1,549 @@
...
+- [Tagged](#tagged)
+
+References:
+
+- [Target](#target)
+- [Promise](#promise)
+
+# Atom
+
+Atoms are values that do not contain or refer to other values.
```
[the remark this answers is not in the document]
kriskowal: No labyrinth can contain you.

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: I am now committing an editorial change to both String and Selector that maintains "surrogate" (implied code point, and as such, pairing code units is not a thing, and expanded the non-normative explanation and implication for unpaired surrogate code units)

draft-specifications/Model.md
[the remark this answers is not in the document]
kriskowal: Electing to omit since these are examples and not exhaustive. May revisit JSON holistically.

draft-specifications/Model.md
```diff
@@ -0,0 +1,613 @@
...
+
+> - **Guile**: `-1`, `0`, `1`
+> - **JavaScript**: `-1n`, `0n`, `1n`
+> - **Python**: `-1`, `0`, `1`
+>
+> Note: We achieved consensus on the name `Integer` at the [November 14, 2023
+> meeting](https://github.com/ocapn/ocapn/issues/94).
+
+For purposes of [Pass Invariant Equality](#pass-invariant-equality), every
+Integer value is only equal to other Integers with the same magnitude.
```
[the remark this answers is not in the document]
kriskowal: Adopted markm’s recommended “same arithmetic integer” prose instead.

kriskowal:
Posted a fresh draft, now pending @tsyesika’s approval only.

gibson042 (approved):
Just a few remaining nits.

tsyesika (approved):

PR #155, opened 2025-03-04
Document abstract notation and concrete representation of the data model

kriskowal:


draft-specifications/Model.md
```diff
@@ -84,6 +98,13 @@ there is only one Null value and it is equal to itself.
 
 A value that is either true or false.
 
+The concrete representation of a Boolean is a bare
+[Boolean](Notation.md#boolean), simply `f` or `t`.
+
+```
+<'null>
+```
+
```
kriskowal, line 107:
```suggestion
```

draft-specifications/Model.md
```diff
@@ -113,6 +133,9 @@ arithmetic integer.
 
 An IEEE 754 64-bit floating point number.
 
+Float64 is represented by [Float64](Notation.md#float64) in the underlying
+abstract notation and concrete representation.
```
kriskowal, line 137: Canonical NaN?

draft-specifications/Model.md
```diff
@@ -326,6 +375,23 @@ Each key must be a [String](#string), and must be
 non-[Equal](#pass-invariant-equality) to any other key within a Struct.
 [Values](#value) within a Struct may be of heterogeneous type.
 
+The abstract notation for a Struct is a [Struct](Notation.md#struct) with the
+abstract notations of the respective keys and values.
+For example, the abstract notation for a Struct with a field named `"a"` and an
+Integer value of `1` is:
```
kriskowal, line 381: Canonical field order?

draft-specifications/Model.md
```diff
@@ -298,6 +330,23 @@ A container is a value that contains other values.
...
+```
+[ 1 "a" ]
+```
+
+Likewise, the concrete representation of a List is a [List](Notation.md) of the
+representations of the respective contained values.
+The concrete representation of the same list is:
+
+```
+[ 1+ 1"a ]
```
kriskowal, line 347: Canonicalize (remove space)?

kriskowal (commented):
We have agreed at the March 11, 2025 meeting to lean on a canonicalized concrete representation, wire format, serialized bytes, including specifying the canonical NaN, canonical strings, no spaces, bytewise ordered field keys.

Names of layers are confusing. Contenders:
- abstract data model
- presentation format
- wire format, binary, serialization, encoding, codec, serdeser

kriskowal:
I’ve posted an update that adopts the terms Presentation Format and Wire Format, for @tsyesika’s consideration.

I’ve also moved to make the Wire Format strictly canonical.

There remains an open question about Structs and the range of field-names. As I’ve written, if the field-names are of heterogeneous type, the Wire Format orders them according to the bytewise representation in the Wire Format of each field-name, which makes the order sensitive to the Wire Format. For Data, this is unimportant because we restrict the range of valid field names to Wire Format Strings.

If we restricted Structs to string field names throughout CapTP, we could simplify the specification and make this weird detail irrelevant. I don’t believe the CapTP specification makes any use of Structs.

So, should we simplify the Wire Format specification further?

erights:
I think we've already decided that the fields of Structs only have string names. In any case, for us to support ocapn from our JS bindings, we need that. I do not think this is a possible degree of freedom.

erights:
>  I don’t believe the CapTP specification makes any use of Structs.

I must be totally misunderstanding then. What about the Structs in the ocapn abstract data model? What am I missing?

kriskowal:
> > I don’t believe the CapTP specification makes any use of Structs.
> 
> I must be totally misunderstanding then. What about the Structs in the ocapn abstract data model? What am I missing?

I’m referring to specific documents: https://github.com/ocapn/ocapn/tree/main/draft-specifications

OCapN consists of a CapTP Specification, Model (passable types), and will eventually include at least one Netlayer. It’s well-established that the Model Struct is a Format Struct and the keys or field names are strictly represented as Model/Format Strings.

You will note that many types in the Model use Format Records, and many Format Records don’t correspond to Data. So there’s not a 1:1 correspondence.

The CapTP Specification covers all Message types and makes extensive use of (Format) Records that do not correspond to data. It _could_ also use Format Structs that do not correspond to passable data, and considering how freely the CapTP uses Format Selector, I’d imagine those would be Structs with Selector field names (not necessarily, I am borrowing an assumption about Spritely’s favored aesthetics.

In JavaScript, my expectation is that we would only ever reïfy types from the Model and we’d use some other form of pattern matching to extract Messages from CapTP, since it makes extensive use of Format Selector and reïfying those as Model Selectors, for which we’d use objects, would put a lot of pressure on our GC.

kriskowal:
> In JavaScript, my expectation is that we would only ever reïfy types from the Model and we’d use some other form of pattern matching to extract Messages from CapTP, since it makes extensive use of Format Selector and reïfying those as Model Selectors, for which we’d use objects, would put a lot of pressure on our GC.

Which is to say, I don’t think Agoric/Endo would use the OCapN format the same way Spritely uses Syrup and Agoric uses JSON today. We’d use a lazy Wire Format parser at the Message boundary and only eagerly reïfy Model data.

draft-specifications/Formats.md
```diff
@@ -0,0 +1,248 @@
...
+- _presentation-character_: _any printable ASCII character except `"` or `\`_ ;
+  note that spaces are printable
+- _wire-string_: _length_ `"` _bytes_
+- _length_: _integer-digits_ ; The number of bytes in _bytes_ as ASCII decimal
+  digits.
+- _bytes_: _byte_* ; The bytes of the string in UTF-8 encoding.
+
+We do not attempt to capture strings with embedded quotes or non-ASCII Unicode
+characters in the Presentation Format, but all Unicode strings excluding
+surrogates (U+D800-U+DFFF) are expressible in the Wire Format.
```
erights, line 142: Since you designed the Presentation Format to not be able to express everything in the model nor everything in the Wire Format, I do not understand the purpose of the Presentation Format. 

For example, the string consisting of a double quote character must be in both the Model and the Wire Format. What can the Presentation Format be useful for if it simply cannot represent this string?

I would have expected anything in the Model to round trip fine between Presentation Format and Wire Format. But the Presentation Format is optimized for human readability / writability and is not canonical. What am I missing?

Strings are not the only place this seems to arise.

erights: Also, surely we specify that the wire format uses UTF-8. But it isn't mentioned here, which I also do not understand.

kriskowal: The Presentation Format is a spec fiction. The specification only obliges an implementation to implement the Wire Format. The purpose of the Presentation Format is to simplify the writing of the specification.

Consider the example at https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#promise-pipelining-by-example

```
<op:deliver <desc:export 5> ['make-car-factory] 3 <desc:import-object 15>>
<op:deliver <desc:answer 3> ['make-car] 4 <desc:import-object 16>>
<op:deliver <desc:answer 4> ['drive] 5 <desc:import-object 17>>
```

This example is written in the Presentation Format. Maintaining the specification in Wire Format is possible but tedious. The corresponding Wire Format for the first line is:

```
<10'op:deliver<11'desc:export5+[16'make-car-factory]3+<18'desc:import-object15+>>
...
```

kriskowal: To be clear, we _could_ take the time to fully specify a many:one mapping from presentation to wire and one:many from wire to presentation for all expressible messages, but I did not deem it worth the time. Preserves and Syrup are certainly exhaustive.

draft-specifications/Formats.md
```diff
@@ -0,0 +1,248 @@
...
+
+> Examples:
+> - `{ a: 10, b: 2 }` corresponds to `{1"a10+1"b2+}`.
+> - `{ "a": 10, "b": 2 }` corresponds to `{1"a10+1"b2+}`.
+> - `{ 'b: 2, 'a: 10 }` corresponds to `{1'a10+1'b2+}`.
+
+- _presentation-struct_: `{` ( _presentation-field_ ( `,` _presentation-field_ )* )? `}`
+- _presentation-field_: _presentation-key_ `:` _presentation-value_
+- _presentation-key_: _presentation-field-name_ / _presentation-value_
+- _presentation-field-name_: _name_ ; Corresponding to a string.
```
erights, line 197: Assuming we come to agreement that this should remain "Corresponding to a string", the String presentation format has an initial and terminal double quote, which explains the second example line above. I would like to allow the first example line, even at the cost of defining a presentation-identifier or something.

Allowing the third example line, for a presentation format for an interoperable protocol seems weird, obscure, and unnecessary. But until I understand better what purpose you have in mind for the Presentation Format, I may be confused about this as well.

draft-specifications/Formats.md
```diff
@@ -0,0 +1,248 @@
...
+- _wire-struct_: `{` _wire-field_* `}`
+- _wire-field_: _wire-key_ _wire-value_
+- _wire-key_: _wire-value_ ; The Wire Format lacks an analog for the
+  Presentation Format's string key shorthand.
+
+> The [Model](Model.md) limits field names in structs to strings, but for
+> purposes of [CapTP](CapTP%20Specification.md) surrounding data, the notation
+> and representation allow any value.
```
erights, line 205: > the notation and representation allow any value.

Is "the notation" the presentation format?

Is "the representation" the wire format?

draft-specifications/Formats.md
erights: Nevermind my question about the first example line. I see you dealt with it below. We therefore need to define presentation-identifier or something.

draft-specifications/Formats.md
kriskowal: I missed one in the translation. Thank you for catching it. Indeed, [Abstract] Notation became Presentation Format and [Concrete] Representation became Wire Format.

draft-specifications/Formats.md
```diff
@@ -0,0 +1,248 @@
...
+surrogates (U+D800-U+DFFF) are expressible in the Wire Format.
+
+## Selector
+
+A sequence of Unicode code points excluding surrogates (U+D800-U+DFFF).
+
+> Example: `'fleur-de-lis` corresponds to `12'fleur-de-lis`.
+
+- _presentation-selector_: `'` _name_
+- _name_: _alpha_ ( _alpha_ / _digit_ / `-` / `:` )*
```
kriskowal, line 151: @erights This is the _identifier_ you seek.

draft-specifications/Model.md
```diff
@@ -344,16 +406,33 @@ non-[Equal](#pass-invariant-equality) to any other key within a Struct.
...
 
 A pair consisting of a tag and a [Value](#value).
-The tag must be a [String](#string).
+
+The Presentation Format for a Tagged value is a [Record](Formats.md#record)
+with three values.
+The first value is a [Selector](Formats.md#selector) spelled "desc:tagged".
+The second value is a [Selector](Formats.md#selector) containing the tag.
+The third value is the Presentation Format of the tagged value itself.
+For example, the Presentation Format for an Integer value of 42 tagged
```
erights, line 424:
```suggestion
The third value is the Presentation Format of the tagged payload itself.
For example, the Presentation Format for an Integer payload of 42 tagged
```

draft-specifications/Model.md
```diff
@@ -407,6 +486,22 @@ A local target handles deliveries and produces either a return value
...
+values.
+The first value is a [Selector](Formats.md#selector) indicating whether the
+value corresponds to a local or remote target.
+The second value is a positive [Integer](Formats.md#integer) specifying a
+target as tracked by the sender or receiver.
+
+If the Selector is spelled `desc:import-object`, the target is local to the
+sender and remote to the receiver of the value.
+If the Selector is spelled `desc:export`, the target is remote to the
+sender and local to the receiver of the value.
```
erights, line 499: What about the three party case, where the reference to Carol in VatC is remote to both Alice in VatA and Bob in VatB. What ocapn does vatA send to vatB to talk about the mutually remote target?

kriskowal: My understanding is that the transfer occurs out-of-band, but I have not attempted to implement 3PH based on the CapTP spec myself and so defer to @tsyesika. Does the in-band data model require special records for 3PH or will these suffice? https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#third-party-handoffs

draft-specifications/Formats.md
erights: The purpose of a fully specified, round-trippable presentation format is that it can be read and written by humans clearly, and so can be used as a notation within programs for values in the model. It can be compiled into the wire format, and the wire format can be parsed into the presentation format.

What about non-canonical smallcaps as a fully specified, round-trippable presentation format, i.e., smallcaps with the normal JSON degrees of freedom: whitespace and order of fields? Then, typically the program manipulable form will be the tree of plain json objects that trivially round trip to the presentation format via `JSON.parse` and `JSON.stringify`. Another pleasant consequence is that it can all be processed by jq and similar generic JSON-based tools.

kriskowal: The less invasive alternative would be to absorb more of [Preserves](https://preserves.dev/cheatsheet.html) into the Presentation Format. The only parts we need that I’ve omitted for expedience are string and symbol quotes and escapes.

With @tsyesika’s blessing, I would not mind omitting the shorthands for struct field names and the first symbol of a record.

kriskowal:
@tsyesika I’m taking it as a given from @erights feedback above that it’s a requirement on our side that the Presentation Format used throughout the CapTP spec needs to be able to express every expressible message in Wire Format, and that every expressible message in Wire Format must have at least one representation in PresentationFormat. So, I’ll follow up with a change that brings in more of Preserves notation.

I’m noticing at this late stage that Preserves would put single quotes on both sides of a Selector, so I’ll follow suit.

I’m also noticing that Preserves is very liberal about what code points may appear in Strings and Selectors. I’ll roll with it, but I don't love that his is a valid Preserves String.

```
"
              👻"
```

I’m also unclear about what to do about comment introducers that appear inside there.

```
"
    ; comment?
"
```

I would not mind if Presentation Format (unlike Preserves) were limited to printable ASCII with escapes for arbitrary code points.

I’m going to remove the conceit that a field names will tend to be identifiers. Like JSON, let’s always double-quote Struct Field-Names. That will make the grammar easier, particularly around arbitrary Unicode.

I’ll also reword ByteArray to look more like Preserves.

Issue #163, opened 2025-04-24
Syrup format for "public-key" data structure not specified


kumavis:
specification only provides this
https://github.com/ocapn/ocapn/blob/3bb8575a9b6a6de481c126cd0d1650ce48a30f68/draft-specifications/CapTP%20Specification.md?plain=1#L261

the [python test-suite](https://github.com/ocapn/ocapn-test-suite) serializes it as:
```syrup
[ 'public-key [ 'ecc [ 'curve 'Ed25519 ] [ 'flags 'eddsa ] [ 'q <bytestring> ] ] ]
```

kumavis:
notably there is no `s` field

kumavis:
On first reading of the spec I didn't realize "syrup sequence" was referring specifically to a Syrup [Sequence](https://github.com/ocapn/syrup/blob/2214cbb7c0ee081699fdef64edbc2444af2bb1d2/draft-specification.md#sequences), related to a OCapN data model [List](https://github.com/ocapn/ocapn/blob/3bb8575a9b6a6de481c126cd0d1650ce48a30f68/draft-specifications/Model.md#list)

This would be easier to follow if we used the **syrup abstract syntax** here instead of the s-expression syntax

PR #166, opened 2025-04-28
Fix: improve handoff sections + cryptography format spec + fix additional documents links

kumavis:
Large improvement:
- Improves clarity of third party handoff sections

Smaller changes:
- Fixes Public Key format spec (there should be no `s`-value)
- Specify cryptography formats using "presentation format"
- Fix broken links in "Additional Documents" section

Resolves https://github.com/ocapn/ocapn/issues/174
Resolves https://github.com/ocapn/ocapn/issues/163

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -324,7 +324,7 @@ imported from a different session, a handoff MUST occur. The handoff is
 initated by the Gifter doing two things:
 
 - Depositing a "gift" to the exporter's bootstrap object.
-- Creating and sending a signed `desc:handoff-give` to the Recipient in place of the [Reference][Model-Reference].
+- Creating and sending a signed `desc:handoff-give` to the Receiver in place of the [Reference][Model-Reference].
 
 ### [Depositing the gift](#depositing-the-gift)
 
@@ -449,10 +449,10 @@ This should have been sent with the [`op:deliver-only`](#op-deliver-only) operat
 Here is an example of how to use this method:
 
 ```text
-<op:deliver-only <desc:export 0>         ; Remote bootstrap object
-                 ['deposit-gift          ; Argument 1: Symbol "deposit-gift"
-                  gift-id                ; Argument 2: Non-negative integer (>=0)
-                  <desc:import-object]>  ; Argument 3
+<op:deliver-only <desc:export 0>            ; Remote bootstrap object
+                 ['deposit-gift             ; Argument 1: Symbol "deposit-gift"
+                  gift-id                   ; Argument 2: Non-negative integer (>=0)
+                  <desc:import-object 5>]>  ; Argument 3: object being shared via handoff
 ```
 
 ## `withdraw-gift` Method
@@ -470,7 +470,7 @@ Here is an example of how to use this method:
 ```text
 <op:deliver <desc:export 0>           ; Remote bootstrap object
             [withdraw-gift            ; Argument 1: Symbol "withdraw-gift"
-             <desc:handoff-receive>]  ; Argument 2: desc:handoff-receive
+             <desc:handoff-receive>]  ; Argument 2: desc:handoff-receive (wrapped in sig:envelope)
             1                         ; Answer position: Non-negative integer (>=0)
             <desc:import-object 3>>   ; The object exported (by us) at position 3, should receive the gift.
 ```
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -258,12 +258,11 @@ public keys and the SHA512 hash algorithm. The EdDSA public keys are based on
 the Ed25519 elliptic curve. The public key is formatted as follows:
 
 ```text
-(public-key (ecc (curve Ed25519) (flags eddsa) (q ...) (s ...)))
+['public-key ['ecc ['curve 'Ed25519] ['flags 'eddsa] ['q q_value]]]
 ```
 
-In the above format, each `(` and `)` mark the start and end of a new Syrup
-sequence with each value with the sequence being a symbol except the `...`,
-which is replaced by Binary Data typed values representing the public key.
+In the above format, the `q_value` is a [ByteArray](https://github.com/ocapn/ocapn/blob/3bb8575a9b6a6de481c126cd0d1650ce48a30f68/draft-specifications/Model.md#bytearray) value of 32 bytes,
+representing the public key.
 
 ## [Signature](#signature)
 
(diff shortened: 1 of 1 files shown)


draft-specifications/CapTP Specification.md
```diff
@@ -258,25 +258,23 @@ public keys and the SHA512 hash algorithm. The EdDSA public keys are based on
 the Ed25519 elliptic curve. The public key is formatted as follows:
 
 ```text
-(public-key (ecc (curve Ed25519) (flags eddsa) (q ...) (s ...)))
+['public-key ['ecc ['curve 'Ed25519] ['flags 'eddsa] ['q q_value]]]
```
kriskowal, line 261: Our spec’s “presentation format” (what I was calling “abstract notation”) doesn’t yet have an explicit affordance for interpolated variables like `q_value`, but as far as I’m concerned, this is fine. We can back-fill bare identifiers, as long as they are not possible to confuse for a shorthand notation on symbols.

kriskowal:
Leaving this review to our Spritely editor, @tsyesika, because I do not know whether the intent is to model keys as strings or embedded Syrup. I gather that it’s the latter.

draft-specifications/CapTP Specification.md
```diff
@@ -290,99 +307,100 @@ fashion, even if the messages are viewed by a malicious actor.
...
+imported from a different session, a handoff MUST occur. The handoff is
+initated by the Gifter doing two things:
 
-and:
-1.  Creating and sending a [`desc:handoff-give`](#desc-handoff-give)
-2.  Signing the `desc:handoff-give` and wrapping it within a `desc:sig-envolope`
-3.  Replace the reference to the remote object being gifted in the message to be
-    the signed handoff give in step 2
+- Depositing a "gift" to the exporter's bootstrap object.
+- Creating and sending a signed `desc:handoff-give` to the Recipient in place of the [Reference][Model-Reference].
```
tsyesika, line 327: Maybe this should use receiver, rather than recipient

draft-specifications/CapTP Specification.md
```diff
@@ -419,21 +437,22 @@ An example of how to use this method is:
...
 
 ```text
-<op:deliver-only <desc:export 0>                               ; Remote bootstrap object
-                 ['deposit-gift                                ; Argument 1: Symbol "deposit-gift"
-                  gift-id                                      ; Argument 2: Non-negative integer (>=0)
-                  <desc:import-object | desc:import-promise>]> ; Argument 3
+<op:deliver-only <desc:export 0>         ; Remote bootstrap object
+                 ['deposit-gift          ; Argument 1: Symbol "deposit-gift"
+                  gift-id                ; Argument 2: Non-negative integer (>=0)
+                  <desc:import-object]>  ; Argument 3
```
tsyesika, line 455: This is missing the `>` for <desc:import-object>, it also probably wants to include a specific imported object e.g. `<desc:import-object 5>` or something.

I think it's fine if we don't, but while we're improving the spec's wording I wonder if we should include a description of this argument, e.g. `Argument 3: object being shared via handoff`.

tsyesika (commented):
Thanks for your work cleaning up the spec. I have left two small comments, otherwise this looks good to merge.

draft-specifications/CapTP Specification.md
kumavis: good catch thanks

draft-specifications/CapTP Specification.md
```diff
@@ -47,11 +47,11 @@ which together build up OCapN (Object Capability Network) specifications.
 
 This specification uses the following other specifications:
 
--   [Syrup](): The serialization format used for all messages between actors
+-   [Syrup](https://github.com/ocapn/syrup): The serialization format used for all messages between actors
```
kriskowal, line 50: Might be worth linking our specific subset in Notation.md (or perhaps Notation.md is confusing and should be cut.)

draft-specifications/CapTP Specification.md
```diff
@@ -443,15 +462,15 @@ in order to receive a gift. It has one argument:
...
+This should have been sent with the [`op:deliver`](#op-deliver) operation, the response the
 bootstrap object should give is the gift which was (or will be) deposited.
 
 Here is an example of how to use this method:
 
 ```text
 <op:deliver <desc:export 0>           ; Remote bootstrap object
             [withdraw-gift            ; Argument 1: Symbol "withdraw-gift"
-             <desc:handoff-receive>]  ; Argument 2: desc:handoff-receive
+             <desc:handoff-receive>]  ; Argument 2: desc:handoff-receive (wrapped in sig:envelope)
```
kriskowal, line 473:
```suggestion
             <desc:handoff-receive>]  ; Argument 2: sig:envelope containing desc:handoff-receive
```

draft-specifications/CapTP Specification.md
```diff
@@ -770,143 +789,136 @@ result.
...
-the [Third Party Handoffs section](#third-party-handoffs) first.
-
-This certificate is provided by the gifter to the exporter, the exporter then
-creates a `desc:handoff-receive` which includes this `desc:handoff-give`
-certificate. The point of this certificate is to provide the exporter some
-crucial information that it needs to check the `desc:handoff-receive` message.
-
-How to check the validity of this certificate is covered in the
-[`desc:handoff-receive`](#desc-handoff-receive).
+This record is used in [Third Party Handoffs](#third-party-handoffs).
```
kriskowal, line 792: There’s a lot of passive voice throughout this document as it stands, but case by case, it’s worth asking whether you can inject the actor into the sentence.
```suggestion
In [Third Party Handoffs](#third-party-handoffs), the Gifter creates a `desc:handoff-give` and sends it to the Receiver.
The Receiver uses the `desc:handoff-give` to redeem the gift from the Exporter.
```

draft-specifications/CapTP Specification.md
```diff
@@ -770,143 +789,136 @@ result.
...
+This record is a certificate created by the Gifter, sent to the Recevier,
+and ultimately used by the Receiver to redeem the gift from the Exporter.
 
-This message MUST always be encapsulated in a
-[`desc:sig-envelope`](#desc-sig-envelope) with a valid signature.
+This record is prepared in the following way:
```
kriskowal, line 796:
```suggestion
The Gifter prepares the record by,
```

kriskowal (approved):
Thank you.

For those following on, the CapTP specification does not yet reflect consensus. With this stamp, I am approving editorial improvements with the expectation that this will ease further consideration.

Issue #168, opened 2025-05-12
Bug: no way to describe an imported answer over the wire


kumavis:
### background

An **AnswerPromise** is a normal Promise, but is only created in response to an `op:deliver` message with a non-`false` `answer-pos` field). The id is specified by the sender to enable Promise Pipelining, but it is otherwise a normal Promise local to the `op:deliver` message receiver.

### problem

In the `op:deliver` message `args` field you specify a List of [Values](https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md#value), potentially including [References](https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md#reference-capability).

When specifying References that are local to the receiver, your options are `desc:export` for Promise and Target, or `desc:answer` for an **AnswerPromise**. 

When specifying References remote to the receiver, your options are `desc:import-promise` for Promise and `desc:import-object` for Target. 
This precludes one type of transmittable Reference: an **AnswerPromise** remote to the receiver.

<table>
  <tr>
    <th></th>
    <th>Target</th>
    <th>Promise</th>
    <th>AnswerPromise</th>
  </tr>
  <tr>
    <th>Local to Recipient</th>
    <td>desc:export</td>
    <td>desc:export</td>
    <td>desc:answer</td>
  </tr>
  <tr>
    <th>Remote to Recipient</th>
    <td>desc:import-object</td>
    <td>desc:import-promise</td>
    <td>( ineffable )</td>
  </tr>
</table>

### endo captp slots

In `@endo/captp` we use a `slot` to index a Reference, remote or local. A slot is a tuple of `(type, isRemoteToReceiver, index)`. Here's how the slots are used in the same situations (with the id to 1):

<table>
  <tr>
    <th></th>
    <th>Target</th>
    <th>Promise</th>
    <th>AnswerPromise</th>
  </tr>
  <tr>
    <th>Local to Recipient</th>
    <td>o+1</td>
    <td>p+1</td>
    <td>q+1</td>
  </tr>
  <tr>
    <th>Remote to Recipient</th>
    <td>o-1</td>
    <td>p-1</td>
    <td>q-1</td>
  </tr>
</table>

### Inducing the bug

The bug is not encountered easily because AnswerPromises are not usually transmitted to whom they are remote to. This is because a reference to the remote AnswerPromise is automatically created by calling `op:deliver` with a non-`false` `answer-pos` field, so it does not need to be transmitted in the normal case.

[...]

Issue #169, opened 2025-05-13
Proposal: Descriptor for imported AnswerPromise


kumavis:
### abstract

introduce `desc:import-answer` for describing remote AnswerPromises and allow `desc:import-answer` in all places where `desc:import-promise` is allowed, such as [**Third Party Handoffs**](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#third-party-handoffs).

### motivation

fix for https://github.com/ocapn/ocapn/issues/168

### current spec 

(no `desc:import-answer`)

### proposed spec

Add `desc:import-answer` to [Descriptors](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#descriptors) section
```
<desc:import-answer answer-pos> ; position: positive integer
```

**Third Party Handoffs** can transmit a `desc:import-promise` and it is proposed that this also could be an `desc:import-answer`.

erights:
Given the proposed meaning, this is a question descriptor, not an answer descriptor.

In any case, based on my sense of the mtg today, I suggest this be closed as unplanned.

PR #171, opened 2025-05-13
Add desc:import-answer

kriskowal:
Closes: #168


kumavis:
this updates the spec per proposal at https://github.com/ocapn/ocapn/issues/169

kumavis:
> This proposal was written with the assumption that OCapN attempts to provide Promise Reference Equality, but it does not

https://github.com/ocapn/ocapn/issues/169#issuecomment-3349739161

I recommend we close this PR

PR #176, opened 2025-05-29
Align anchors with Github generated anchors

kriskowal:
Github omits colons rather than translate to dash.


kriskowal:
Notably, Github also creates its own anchors for every heading. Making heading links might be superfluous depending on the intended rendering medium.

tsyesika (approved):
looks like it needs rebasing but otherwise good.

PR #220, opened 2025-12-03
Fix ocapn-peer field ordering and hints are structs

### LINKED ISSUE #216: Spec Ambiguity: ocapn-peer Record value type "hashmap" is underspecified
https://github.com/ocapn/ocapn/blob/main/draft-specifications/Locators.md#syrup-serialization
```
<ocapn-peer designator  ; string
            transport   ; symbol (cannot contain ".")
            hints>      ; hashmap | false
```

What are the valid types for the values?

tsyesika:
The OCapN test suite and Goblins has the `transport` and `designator` fields swapped (i.e. transport first, designator then hints). This brings the spec in line with the test suite. The hints should also be an OCapN struct not a hashmap with keys that are symbols.

Also fixed a couple of typos/spelling mistakes I noticed.

Fix #216

Side note: I think the feeling at Spritely is still that struct is a bit of a misleading name for what is a stringly keyed hashmap. This change does however uses it as it's the current agreed terminology for this data type.

### DIFF
--- a/draft-specifications/Locators.md
+++ b/draft-specifications/Locators.md
@@ -55,22 +55,21 @@ peer.
 
 ## [Syrup Serialization](#peer-syrup-serialization)
 
-It's encoded as a record with the label `ocapn-peer` and three arguments:
+It's encoded as a record with the label `ocapn-peer` (symbol) and three
+arguments:
 
 ```
-<ocapn-peer designator  ; string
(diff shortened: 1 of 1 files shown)


kumavis:
the struct typing could use some more refinement because the values can only be strings, but not blocking this PR on it

PR #221, opened 2025-12-03
Specify protections against multiple session to the same peer

### LINKED ISSUE #214: Spec needed: Redundant op:session-start on same connection for established session
Unrelated to "crossed-hellos problem", a misbehaving client might send two `op:session-start` messages **on the same connection.**

The spec states:
> It's important that we only have one bidirectional connection between a CapTP session.

However it does not cover when a redundant `op:start-session` is sent on a existing connection.

In my eyes, the only correct response is to abort the session.

tsyesika:
Since there can be only one session between two peers, there spec needs to cover how we achieve this. It was under-specified and unclear.

This change improves both why we only want one session between the same two peers and that we need to check:

- It doesn't have a connection already when wanting to establish a new one.
- If multiple `op:start-sessions` are received the session is terminated.
- Abort newly received connections when there already is a valid active session.
- Specify how to detect and mitigate crossed hellos more thoroughly.

Fix #214

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -261,12 +261,12 @@ the Ed25519 elliptic curve. The public key is formatted as follows:
 ['public-key ['ecc ['curve 'Ed25519] ['flags 'eddsa] ['q q_value]]]
 ```
 
-In the above format, the `q_value` is a [ByteArray][Model-ByteArray] value of 32 bytes,
-representing the public key.
+In the above format, the `q_value` is a [ByteArray][Model-ByteArray] value of
+32 bytes, representing the public key.
 
-## [Public ID](#public-id)
+## [Public Identifier](#public-identifier)
 
-The Public ID for a peer is a [ByteArray][Model-ByteArray] of length 32.
+The Public Identifier for a peer is a [ByteArray][Model-ByteArray] of length 32.
 
 1. Serialize the per session public key [as described here](#public-key).
 2. SHA256 hash of the result produced in step 1.
@@ -276,9 +276,9 @@ The Public ID for a peer is a [ByteArray][Model-ByteArray] of length 32.
 
 The Session ID for a session is a [ByteArray][Model-ByteArray] of length 32.
 
-1. Calculate the Public ID of each side using [the process described here](#public-id).
+1. Calculate the Public Identifier of each side using [the process described here](#public-identifier).
 2. Sort both IDs based on the resulting octets.
-3. Concatinate the Public IDs in the order determined in step 2.
+3. Concatinate the Public Identifiers in the order determined in step 2.
 4. Prepend the string "prot0" to the beginning.
 5. SHA256 hash the result from step 4.
 6. SHA256 hash the result from step 5.
@@ -479,8 +479,7 @@ Here is an example of how to use this method:
 
 ## [`op:start-session`](#opstart-session)
 
-When setting up a new session, a EdDSA key pair with a SHA512 hash should be
-generated.
+When setting up a new session, a EdDSA key pair should be generated.
 
 This operation is used when a new session is initiated over CapTP. Both
 parties MUST send an `op:start-session` message upon a new session. The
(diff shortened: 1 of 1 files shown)


draft-specifications/CapTP Specification.md
```diff
@@ -492,13 +492,16 @@ parties MUST send upon a new connection. The operation looks like this:
...
 
-It's important that we only have one bidirectional connection between a CapTP
-session. Before trying to connect to a peer, it's important to check that a
-connection is not already open before proceeding. There is a race condition
-called the "crossed hellos" problem whereby each side tries to open a new
-connection at the same time and only one of these connection attempts must
-succeed. The way to check which attempt should succeed is to sort the serialized
-`session-pubkey`, the highest of the two wins.
+An important aspect of CapTP is that only one active session between two peers
+should exist. This allows peers to equality checks against objects from a remote
```
davexunit: Missing a verb here. Did you mean "This allows peers to *perform* equality checks..."?

draft-specifications/CapTP Specification.md
```diff
@@ -492,13 +492,16 @@ parties MUST send upon a new connection. The operation looks like this:
...
+peer and prevents [3rd Party Handoffs](#third-party-handoffs) from where the
+receiver and exporter are on the same peer.
+
+There are several mechanisms put in place to ensure that only one session exists
+between two peers. The first is that a CapTP implementation MUST check if it has
+an active session with a given peer, before trying to establish a new
+connection. If the peer already has an active session, that session MUST be used
+instead of creating a new one.
```
davexunit: Out of curiosity, does the spec ever explain what "session with a given peer" means exactly? For example, if you have two ways of reaching a peer over two different netlayers, we consider those two different peers, right? I feel like this is a question that someone reading/implementing OCapN would have.

draft-specifications/CapTP Specification.md
```diff
@@ -518,12 +521,45 @@ per-session key pair. This is serialized in accordance with
...
+
+### Crossed Hellos Mitigation
+
+Crossed hellos is a race condition which occurs when two peers attempt to open a
+new session to one another simultaneously. If both sessions were to succeed
+the result would be multiple sessions between the same peers. Since this is
+not permitted in CapTP, all implementations are responsible for detecting and
+mitigating the problem.
+
+Implementations are responsible for keeping track the sessions they'v
```
davexunit: Typo: "they'v"

draft-specifications/CapTP Specification.md
tsyesika: It doesn't. Since we don't currently support using multiple netlayers together, the union of the `transport` and `designator` is the deciding factor. This is an underspecified part of the spec and I suspect we do want to support multiple netlayers which probably requires some global (to all netlayers) set of keys and identity.

draft-specifications/CapTP Specification.md
```diff
@@ -518,12 +522,45 @@ per-session key pair. This is serialized in accordance with
...
+the receiving side should check to see if it has a session it has opened with
+the peer located at the `acceptable-location` provided. If it both received a
+session from a peer that it has also opened a session to, the crossed hellos
+problem has been detected and must be mitigated.
+
+The way to mitigate the problem is by choosing which of the two sessions should
+be allowed to "win" (and the other to be aborted). This is done by
+deterministically calculating the [Public ID](#public-id) for its outbound
+connection and the other side's inbound connection. These two keys (in their
+syrup serialization) should be compared bytewise to each other. The lower of the
```
kriskowal: We should compare the identifiers bytewise directly. The Syrup serialization should not be germane.

kriskowal: I’ll concede that this change makes forward progress and can be merged in advance of consensus, but note that these details are not under consensus.

I need to understand this better. One of my goals is to obtain consensus on an interop target in plenary sooner than later, but for my vote, there has to be a story for making Syrup negotiable in flight. This is a spot where Syrup is getting used too low to be negotiated.

One possible solution is to push more concerns into a network layer. Peer identifiers safely vary between network layers, but cannot safely vary within network layers. So, negotiating the serialization used for constructing the hash of the peer identity for resolving crossed hellos would break the equality invariants within a network.

Another possible solution is to reduce the amount of information captured by a peer identifier. For example, if we forgo hashing and just compare EdDSA public keys, there is no need for a serialization envelope. That does lock the network layer into EdDSA for signing. I had more or less assumed that would be the case regardless.

I’d say there’s no reason to hash the EdDSA except that it discourages, but does not prevent, folks from rolling a a public key for the purpose of tending to win at crossed hellos. I could be educated either way.

draft-specifications/CapTP Specification.md
```diff
@@ -518,12 +522,45 @@ per-session key pair. This is serialized in accordance with
...
 connection MUST be aborted.
 
 The `acceptable-location-sig` MUST be valid that the `session-pubkey` provided a
 valid signature of `acceptable-location`.
 
+The implementation should check if an active valid session already exists
+between the two peers, if one does exist the new session should be aborted.
+
+Detection and (if needed) mitigation of the Crossed Hellos problem described
+below MUST be performed.
```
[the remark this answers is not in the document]
tsyesika: At least at the moment the netlayer is responsible for handling the racing if that's supported by the specific netlayer in question. Crossed hellos resolution only needs the other side to send their `op:start-session`, then each side can independently come to the correct conclusion over which can proceed. This means if one side doesn't implement crossed hellos at all, or has somehow implemented it incorrectly it should still work provided the other side conforms to spec.

draft-specifications/CapTP Specification.md
tsyesika: I think this probably would benefit from being broken into its own issue, I think it's probably substantial enough to be discussed on its own.

With that said there are several places within CapTP where syrup is needed. I'd say the most significant are the 3rd Party Handoff certificates, ultimately you must check multiple signatures and the since syrup is canonical you can easily produce the data that was signed based on the tagged data.

kriskowal: I posted https://github.com/ocapn/ocapn/issues/222

RidleyWrites (approved):
I took a look and all the changes seem fine to me. Sorry for the hold up.

PR #227, opened 2025-12-13
Add Sturdy References to the Model

kriskowal:
I mistakenly omitted Sturdy Reference from the OCapN data model. This change adds them, clarifies their role and semantics, and explicitly calls out the role of a sturdyref locator as the in-band representation of a sturdyref.

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -436,9 +436,8 @@ An example of how to use this method is:
             <desc:import-object 5>>  ; object exported by us at position 5 should provide the answer
 ```
 
-> Fetch does not produce [Sturdy References](./Model.md#sturdy-reference).
-> Fetch produces live references, [Targets](./Model.md#target), for which there
-> may exist a corresponding Sturdy Reference or [Locator](./Locators.md).
+> Fetch may produce any [Value](./Model.md#value) in the model and is the
+> the mechanism for creating original [Sturdy References](./Model.md#sturdy-reference).
 > A mechanism for looking up a Locator or Sturdy Reference for a Target returned
 > by `fetch` is optional, implementation-specific, and necessarilly a closely held
 > capability of the OCapN implementation that should not be revealed to guest
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -437,8 +437,8 @@ An example of how to use this method is:
 ```
 
 > Fetch may produce any [Value](./Model.md#value) in the model and is the
-> the mechanism for creating original [Sturdy References](./Model.md#sturdy-reference).
-> A mechanism for looking up a Locator or Sturdy Reference for a Target returned
+> the mechanism for creating original [Sturdyrefs][Model-Sturdyref].
+> A mechanism for looking up a Locator or Sturdyref for a Target returned
 > by `fetch` is optional, implementation-specific, and necessarilly a closely held
 > capability of the OCapN implementation that should not be revealed to guest
 > code by default.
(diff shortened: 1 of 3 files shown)


draft-specifications/CapTP Specification.md
```diff
@@ -419,11 +419,12 @@ receive both [`op:start-session`](#opstart-session) messages.
 
 ## `fetch` Method
 
-This method is used to fetch an object from the bootstrap object. To use it you
-need a `swiss-number` which is a Binary Data type. This swiss number should
-correspond an object which exists in this session. The result will be the object
-which corresponds to this `swiss-number` or an error if the object does not
-exist or a swiss number was not provided.
+This method is used to fetch a Target object from the bootstrap object.
```
tsyesika, line 422: Does this want a link to Model.md#target ?

draft-specifications/CapTP Specification.md
```diff
@@ -435,6 +436,15 @@ An example of how to use this method is:
...
 ```
 
+> Fetch may produce any [Value](./Model.md#value) in the model and is the
+> the mechanism for creating original [Sturdyrefs][Model-Sturdyref].
+> A mechanism for looking up a Locator or Sturdyref for a Target returned
+> by `fetch` is optional, implementation-specific, and necessarilly a closely held
+> capability of the OCapN implementation that should not be revealed to guest
+> code by default.
+> In E, this capability was called an
+> [Introducer](http://erights.org/elang/concurrency/introducer.html).
```
tsyesika, line 446: A fetch does not create a sturdyref, it's the mechanism to retrieve a reference for a swiss-num. The fetch is also not optional and required by the CapTP specification as part of the bootstrap object. The bootstrap object is available to all peers and exported at position 0.

I think what you're maybe describing is the mechanism to register new objects with and produce swiss-nums to give out. This is not really covered by this spec as it's considered a detail for the implementation and outside the scope of CapTP.

draft-specifications/CapTP Specification.md
```diff
@@ -876,6 +886,28 @@ Party Handoffs](#third-party-handoffs) section.
 <desc:export position>  ; position: positive integer
 ```
 
+## [`ocapn-sturdyref`](#ocapn-sturdyref)
```
tsyesika, line 889: I'm not sure we want to duplicate the definition of a peer and sturdyref when they're already defined in the Locators specification. What do you think?

draft-specifications/Locators.md
```diff
@@ -53,6 +53,7 @@ pieces of information which need to match. Two peer locators can have the same
 designator and transport but  different hints and be considered to be the same
 peer.
 
+<a name="peer-syrup"></a>
 ## [Syrup Serialization](#peer-syrup-serialization)
```
tsyesika, line 57: I know you didn't name this (I did), but I wonder if we should change this to "In band serialisation" opposed to syrup. That can happen in the future PR, just occurred to me when reviewing this.

draft-specifications/Model.md
```diff
@@ -471,6 +477,70 @@ received promises will satisfy the pass invariants applicable to their type.
 > [Containers](#container) maintain [Pass Invariant
 > Equality](#pass-invariant-equality).
 
+## Sturdyref
+
+A Sturdyref (sturdy reference) represents the capability to attempt to obtain a
+live reference (a Target) for a [Locator](./Locator.md), but encapsulating the
+locator.
+
+> - **Guile**: to be proposed
```
tsyesika, line 486: This is defined in Goblins as an `ocapn-sturdyref` record from the model `(goblins ocapn ids)`

draft-specifications/Model.md
```diff
@@ -471,6 +477,70 @@ received promises will satisfy the pass invariants applicable to their type.
...
+>   ({ [Symbol.for('passStyle')]: 'sturdyRef' })
+>   ```
+> - **Python**: to be proposed
+
+OCapN implementations must provide a mechanism that produces a Promise
+for a Target from the encapsulated Locator, by attempting to establish
+a session with the designated Peer and Fetching the designated Swissnum.
+The mechanism should not be closely held.
+That is, the mechanism should require no special capability beyond the Sturdy
+Reference itself.
```
tsyesika, line 500: I think in some implementations (Goblins included at the moment, though there are [discussions](https://codeberg.org/spritely/goblins/issues/267) to potentially change that), enliven is considered a capability that is selectively given out. A sturdyref alone isn't sufficient, you need both the enliven capability and a sturdyref.

I think other implementations may take a different approach and this should probably be left up to individual implementations.

tsyesika (commented):
I think I agree with Ridley. I think it can be a difficult balance providing enough context and info about the main implementations, and having too much non-normative implementation content. I think we're leaning a bit too heavily towards non-normative content at the moment.  We're also assuming one implementation per language at the moment (Goblins being the only Guile implementation, Endo being the only JS implementation) when specifying how things are represented per language.

PR #234, opened 2025-12-13
Correction of Sequence to List in CapTP

kriskowal:
This makes the CapTP specification consistent with the agreement to rename Sequence to List in the Model, and adds cross-references.


tsyesika (approved):

tsyesika:
Approved, I'll leave merging to you @kriskowal

PR #240, opened 2025-12-15
Update Notation text for string/symbol shorthands

gibson042:
Struct field name shorthands must be alphanumeric, but there is no such restriction on a Record discriminator (many of which in https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md include `:` and/or `-`).


kriskowal (approved):
I deem this to be an editorial change beneath plenary review, and if @tsyesika agrees, we can merge this.

jar398:
This looks done, just waiting for @tsyesika

Issue #242, opened 2025-12-15
Why do `deposit-gift` and `op:get`/`op:index`/`op:untag` expect to receive `desc:import-*`?


gibson042:
[CapTP Specification: Descriptors](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#descriptors) is clear about descriptors being interpreted from the receiver's side, such that in messages from Alice to Bob, `desc:import-object` and `desc:import-object` refer to values exported by Alice while `desc:export` and `desc:answer` refer to values exported by Bob.

...which leaves me confused about some other parts of the spec.

For example, [Third Party Handoffs](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#third-party-handoffs) and [`deposit-gift`](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#deposit-gift-method) document that Gifter communicates with Exporter via `deposit-gift` with a `desc:import-object` argument and communicates with Receiver via `desc:handoff-give` in place of some other Value in a message (e.g., that Gifter Alice sends to Exporter Bob's bootstrap object an `op:deliver-only` message with discriminator argument `'deposit-gift` and reference argument identifying a Target exported by Alice, and then sends to Receiver Carol an `op:deliver` message with a `desc:handoff-give` in some argument). But it seems pointless to involve the Exporter at all, since the Gifter could much more directly just use `desc:import-object` with the Receiver.

Likewise with [`op:get`](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#opget), [`op:index`](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#opindex), and [`op:untag`](https://github.com/ocapn/ocapn/blob/main/draft-specifications/CapTP%20Specification.md#opuntag), all of which specify their receiver as `desc:answer | desc:import-promise`. Why would sender Alice ever ask recipient Bob to dive into the structure of a value that Alice is exporting _to_ Bob?

Am I misunderstanding, or are those uses of `desc:import-*` effectively typos that should be replaced with `desc:export`? And further, is there a more general principle whereby any place in which a `desc:import-object` or `desc:import-promise` or `desc:handoff-receive` is valid should accept any of those and any place in which a `desc:export` or `desc:answer` is valid should accept either of those?

RidleyWrites:
I believe you are correct.

[...]

PR #244, opened 2025-12-16
Spec Bug: deposit-gift should specify an export (not an import)

kumavis:
"export" here means an object hosted by the receiver of the message. With this message we're asking this exporter to mark their export as a gift.

noted here https://github.com/ocapn/ocapn/issues/242


RidleyWrites (changes requested):
I think it should permit either a `desc:export` or a `desc:answer`

The change also needs to be made in a second location in the spec:

https://github.com/ocapn/ocapn/blob/23cdd62641e63696c2a66c93b9784e1128bce12a/draft-specifications/CapTP%20Specification.md?plain=1#L346-L346

PR #269, opened 2026-05-12
Add meeting minutes for 2026-05-12

cwebber:


dckc (approved):
seems to me that the June meeting agreed to these minutes, if only implicitly

Issue #167, opened 2025-05-12
Proposal: Single op:deliver message type with resolution subscription via `op:listen`


kumavis:
### abstract

- Remove `resolve-me-desc` from `op:deliver`, replace with (optional) subsequent `op:listen` against `answer-pos`
- Merge `op:deliver-only` and `op:deliver`

### motivation

Creation of a remote promise for the deliver operation should be optional (as with `op:deliver-only`). Subscribing to that remote promise should also be optional (currently not possible).

`op:deliver`'s `resolve-me-desc` and `answer-pos` both deal with the Promise local to the message receiver.
The `answer-pos` is the sender specified id for the Promise local to the receiver. The id is specified by the sender to enable **Promise Pipelining**, but it is otherwise a normal Promise local to the receiver.
If the sender wishes to subscribe to the resolution of the promise created for the deliver operation, there is already a semantic way of doing so: `op:listen`. `op:listen` already supports `desc:answer` as its `to-desc`

To simplify the implementation and specification, `op:deliver-only` should be removed and modeled as an `op:deliver` with `answer-pos` set to `false`.

### current spec

```
<op:deliver to-desc           ; desc:export
            args              ; sequence
            answer-pos        ; positive integer | false
            resolve-me-desc>  ; desc:import-object | desc:import-promise
```

```
<op:deliver-only to-desc  ; desc:export
                 args>    ; Sequence
```

```
<op:listen to-desc           ; desc:export | desc:answer
           listen-desc       ; desc:import-object
           wants-partial?    ; boolean
```

##### `op:deliver`
remove `resolve-me-desc` field:
```
<op:deliver to-desc           ; desc:export
            args              ; sequence
            answer-pos>       ; positive integer | false
``` 

subscribing now optionally performed via `op:deliver` with non-`false` `answer-pos` followed by `op:listen` to `<desc:answer answer-pos>`

##### `op:deliver-only`
entirely remove `op:deliver-only` message type

instead use `op:deliver` with `answer-pos` set to false.

##### `op:listen`
no change.

zarutian:
hmm... having it so that omitting `answer-pos` is equiv setting it to false is further refinement, no?

RidleyWrites:
I've spent some more time thinking about this and trying to explore the tradeoffs.

# Message count tradeoffs

| Positive | - no message for undesired result<br>- no `op:gc-export` for undesired result resolver                                                                          |
[...]

PR #277, opened 2026-06-17
Add op:flush for promise shortening

### LINKED ISSUE #11: Promise Shortening
See https://github.com/Agoric/agoric-sdk/pull/2909#pullrequestreview-638818141

Also, @dtribble please explain the Midori four vat promise shortening case.

### LINKED ISSUE #40: Ordering Guarantees
Forking off of discussion from https://github.com/ocapn/ocapn/issues/5#issuecomment-1487971643:

We'd mostly settled only only providing point to point FIFO ordering guarantees, because the lost resolution bug convinced people that E-Order is apparently harder than originally believed, and it was deemed not worth the trouble. But in light of the way the discussion on issue #5 is going, we may end up building a data model where the lost resolution bug at least cannot cause decoding errors.

With this issue I'd like to understand exactly what the trade-offs are, and what we're buying ourselves by relaxing the ordering requirements. I also want to see if I can align the way capnp does things; having different ordering semantics seems like a great way to have extremely subtle bugs, but right now the capnp docs say we provide E-order. But we also don't do 3PH yet (actively working on it) which is where the hard cases come in, and I'm not sure how wedded we are to that.

Reference: http://erights.org/elib/distrib/captp/WormholeOp.html

I'd like to try to understand what switching to point-to-point FIFO does to the implementation -- what does that look like vs. what E did?

If my understanding is correct, Goblins currently provides E-Order, but *not* the guarantee that references don't unsettle in flight, which the above link seems to suggest is critical for the "Allow services with Near arguments" criterion, so effectively we're dropping that, forcing such services to have to potentially deal with waiting for arguments to resolve, even for things that should end up being near references. I'm currently on the path to implement a similar thing in go-capnp.

Does dropping E-Order lead to a different implementation? If so, what does it look like?

RidleyWrites:
We have been comparing multiple options for message order preservation in #11 and in implementer meetings. So far @gibson042, @kumavis, Dave, and I have all stated a preference for the shortener-initiated solution which is presented in this pull request.

As #11 has gotten rather long and hard to track the various proposals, this PR serves as a reference point for what the favoured proposal actually is.

I can provide an implementation guide update when we have confirmed moving forward with this proposal.

Resolves #11 
Resolves #40

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -131,7 +131,7 @@ should be severed and unresolved promises broken.
 
 Promises are a key part of CapTP. They are used to represent a value which is
 not yet known. Promises without a value are said to be unresolved, they can
-become resolved by being `fulfill`ed with a value 
+become resolved by being fulfilled (`fulfill`) with a value 
 ([including another promise](#promise-shortening)), or broken (`break`) with 
 an error.
 
@@ -409,8 +409,9 @@ preventing unauthorized access to gifts.
 
 # [Promise Shortening](#promise-shortening)
 
-Promise shortening occurs when a promise is fulfilled with a promise on another peer. Extra steps must be 
-taken beyond ordinary reference passing to ensure consistent message ordering.
+When a promise is fulfilled with a promise on another peer, a layer of indirection is removed, thus 
+"shortening" the promise. Extra steps must be taken beyond ordinary reference passing to ensure 
+consistent message ordering.
 
 Promise shortening involves peers with the following roles:
 - **Holder**: The peer holding a reference to a promise hosted by the Shortener that is being given a
@@ -418,8 +419,10 @@ shorter promise on another peer.
 - **Shortener**: The peer hosting a promise that wishes to pass a shorter promise to the Holder.
 - **Terminator**: The peer hosting the promise that the Shortener will be passing to the Holder.
 
-**Note**: It is possible for the Holder and the Terminator to be the same peer.
-**Note**: A Terminator may also play the role of Shortener for a yet-shorter promise on another peer. In that case the Shortener of the prior shortening also plays the role of Holder for the further shortening.
+**Note**: It is possible for the Holder and the Terminator to be the same peer. The same algorithm applies
+in that case as it does when the Holder and Terminator are separate peers.
+**Note**: A Terminator may also play the role of Shortener for a yet-shorter promise on another peer. 
+In that case the Shortener of the prior shortening also plays the role of Holder for the further shortening.
 
 
 ## Promise shortening from the Shortener's perspective
@@ -433,12 +436,12 @@ While waiting for a response to the [`op:flush`](#opflush) message, the Shortene
 
 The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
 new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
-a regular [Third Party Handoff](#third-party-handoffs).
+a [Third Party Handoff](#third-party-handoffs).
 
-**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message it sent
-to the Holder, the Shortener becomes aware of a yet-shorter promise by receiving its own [`op:flush`](#opflush) 
-message. In that case the Shortener MUST wait until it receives the yet-shorter promise and then pass *that*
-promise to the original Holder rather than the promise it originally intended to.
+**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message the
+Shortener sent to the Holder, the Shortener becomes aware of a yet-shorter promise by receiving its own 
+[`op:flush`](#opflush) message. In that case the Shortener MUST wait until it receives the yet-shorter promise 
+and then pass *that* promise to the original Holder rather than the promise it originally intended to.
 
 ## Promise shortening from the Holder's perspective
 
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -131,8 +131,9 @@ should be severed and unresolved promises broken.
 
 Promises are a key part of CapTP. They are used to represent a value which is
 not yet known. Promises without a value are said to be unresolved, they can
-become resolved by being `fulfill`ed with a value (including another promise),
-or broken (`break`) with an error.
+become resolved by being `fulfill`ed with a value 
+([including another promise](#promise-shortening)), or broken (`break`) with 
+an error.
 
 Promises are often created by sending an `op:deliver` message, where they
 represent the eventual value of the response. They can be chained together in
@@ -405,6 +406,53 @@ the **Gifter-Receiver** session. The Exporter will likely not have seen this Pub
 The Receiver should be the only party able to withdraw the gift left by the Gifter.
 Implementers MUST ensure that the management of gifts adheres to this requirement,
 preventing unauthorized access to gifts.
+
+# [Promise Shortening](#promise-shortening)
+
+Promise shortening occurs when a promise is fulfilled with a promise on another peer. Extra steps must be 
+taken beyond ordinary reference passing to ensure consistent message ordering.
+
+Promise shortening involves peers with the following roles:
+- **Holder**: The peer holding a reference to a promise hosted by the Shortener that is being given a
+shorter promise on another peer.
+- **Shortener**: The peer hosting a promise that wishes to pass a shorter promise to the Holder.
+- **Terminator**: The peer hosting the promise that the Shortener will be passing to the Holder.
+
+**Note**: It is possible for the Holder and the Terminator to be the same peer.
+**Note**: A Terminator may also play the role of Shortener for a yet-shorter promise on another peer. In that case the Shortener of the prior shortening also plays the role of Holder for the further shortening.
+
+
+## Promise shortening from the Shortener's perspective
+
+When a Shortener becomes aware that a promise it is sharing to a Holder has been fulfilled with a promise
+on another peer (which could be the Holder), the Shortener sends the Holder an [`op:flush`](#opflush) message
+targeting the listener resolver for the Shortener's promise.
+
+While waiting for a response to the [`op:flush`](#opflush) message, the Shortener MUST continue to forward any
+[promise pipelining](#promise-pipelining) messages received from the Holder to the promise being shortened.
+
+The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
+new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
+a regular [Third Party Handoff](#third-party-handoffs).
+
+**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message it sent
+to the Holder, the Shortener becomes aware of a yet-shorter promise by receiving its own [`op:flush`](#opflush) 
+message. In that case the Shortener MUST wait until it receives the yet-shorter promise and then pass *that*
+promise to the original Holder rather than the promise it originally intended to.
+
+## Promise shortening from the Holder's perspective
+
+When a Holder receives an [`op:flush`](#opflush) message, it creates a new 
+[promise-resolver pair](#promise-objects) and fulfills the target resolver of the [`op:flush`](#opflush) with
+the new promise. As the new promise is a local promise, this prevents any further 
+[promise pipelining](#promise-pipelining) messages being sent through this promise to the Shortener.
+
+The Holder then sends the new resolver to the resolver specified in the Shortener's [`op:flush`](#opflush)
+message.
+
+Later, when the Shortener fulfills the new resolver with the shorter promise reference, any messages sent to
+the Holder's promise since receiving the [`op:flush`](#opflush) message are 
+[promise pipelined](#promise-pipelining) on the shorter promise reference.
 
 # [The bootstrap Object](#bootstrap-object)
 
@@ -697,6 +745,30 @@ An `op:listen` request should NOT be notified when the promise is fulfilled
 with another promise on the same peer unless that promise has been settled to
 either a value or an error, in which case the `op:listen` request is informed
 of the settled result.
+
+## [`op:flush`](#opflush)
+
+This operation is used during [promise shortening](#promise-shortening) to
+preserve message ordering.
+
+```
+<op:flush   to-desc             ;; <desc:export>
+            confirm-desc>       ;; <desc:import-object> | <desc:import-promise>
+```
+
+### Sending
+
+`to-desc` MUST be a `desc:export` which corresponds to a resolver on the remote side.
+
+`confirm-desc` MUST be a `desc:import` representing a resolver on the sender's side
+that will receive the flush confirmation containing a new resolver to send the 
+shortened promise reference.
+
+### Receiving
+
+Upon receiving the `op:flush` message, create a new [promise-resolver pair](#promise-objects).
+Fulfill the target resolver at `to-desc` with the new promise and fulfill the remote resolver
+at `confirm-desc` with the new resolver.
 
 ## [`op:get`](#opget)
 
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -414,8 +414,7 @@ When a promise is fulfilled with a promise on another peer, a layer of indirecti
 consistent message ordering.
 
 Promise shortening involves peers with the following roles:
-- **Holder**: The peer holding a reference to a promise hosted by the Shortener that is being given a
-shorter promise on another peer.
+- **Holder**: The peer holding a reference to a promise hosted by the Shortener.
 - **Shortener**: The peer hosting a promise that wishes to pass a shorter promise to the Holder.
 - **Terminator**: The peer hosting the promise that the Shortener will be passing to the Holder.
 
@@ -429,7 +428,7 @@ In that case the Shortener of the prior shortening also plays the role of Holder
 
 When a Shortener becomes aware that a promise it is sharing to a Holder has been fulfilled with a promise
 on another peer (which could be the Holder), the Shortener sends the Holder an [`op:flush`](#opflush) message
-targeting the listener resolver for the Shortener's promise.
+targeting the listener's resolver for the Shortener's promise.
 
 While waiting for a response to the [`op:flush`](#opflush) message, the Shortener MUST continue to forward any
 [promise pipelining](#promise-pipelining) messages received from the Holder to the promise being shortened.
@@ -443,7 +442,7 @@ Shortener sent to the Holder, the Shortener becomes aware of a yet-shorter promi
 receiving its own [`op:flush`](#opflush) message. In that case the first Shortener MUST continue to forward 
 any messages from the Holder to the second Shortener until the first Shortener receives the flush response 
 from the Holder. Upon receiving the flush response from the Holder, the first Shortener MUST pass the
-reference to the next Shortener, even if it has access to an even shorter promise.
+reference to the second Shortener, even if it has access to an even shorter promise.
 
 While the first Shortener still forwards messages from the Holder to the second Shortener during flushing,
 the first Shortener MUST still prevent any other messages from being sent through to the second Shortener.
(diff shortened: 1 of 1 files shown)


draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,53 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+targeting the listener resolver for the Shortener's promise.
+
+While waiting for a response to the [`op:flush`](#opflush) message, the Shortener MUST continue to forward any
+[promise pipelining](#promise-pipelining) messages received from the Holder to the promise being shortened.
+
+The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
+new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
+a regular [Third Party Handoff](#third-party-handoffs).
+
+**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message it sent
```
davexunit: Is "it" the shortener?

RidleyWrites:
Question: This proposal requires flushing when shortening to a promise on another peer. Is flushing also necessary when fulfilling a promise with an object on another peer? I seem to recall @kumavis raising this concern at one point but might be mixed up.

draft-specifications/CapTP Specification.md
```diff
@@ -698,6 +746,30 @@ with another promise on the same peer unless that promise has been settled to
 either a value or an error, in which case the `op:listen` request is informed
 of the settled result.
 
+## [`op:flush`](#opflush)
+
+This operation is used during [promise shortening](#promise-shortening) to
+preserve message ordering.
```
davexunit, line 755: Might be worth expounding on this somewhere? I think Mark had a phrase "something something FIFO" that I thought was a nice way of describing what OCapN's ordering guarantee ought to be.

draft-specifications/CapTP Specification.md
```diff
@@ -131,8 +131,9 @@ should be severed and unresolved promises broken.
 
 Promises are a key part of CapTP. They are used to represent a value which is
 not yet known. Promises without a value are said to be unresolved, they can
-become resolved by being `fulfill`ed with a value (including another promise),
-or broken (`break`) with an error.
+become resolved by being `fulfill`ed with a value 
+([including another promise](#promise-shortening)), or broken (`break`) with 
+an error.
```
davexunit, line 136: This is a bit tangential but your addition to this sentence made me notice it. It's weird that we use two different formats to describe the resolver methods: "`fullfill`ed" and "broken (`break`)". They should both be the same format like "fulfilled (`fulfill`)" and "broken (`break`)".

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,53 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
 Implementers MUST ensure that the management of gifts adheres to this requirement,
 preventing unauthorized access to gifts.
 
+# [Promise Shortening](#promise-shortening)
+
+Promise shortening occurs when a promise is fulfilled with a promise on another peer. Extra steps must be
```
davexunit: Perhaps it would be worth explaining that the promise on another peer is removing a layer of indirection thus why it is called shortening.

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,53 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+Promise shortening occurs when a promise is fulfilled with a promise on another peer. Extra steps must be 
+taken beyond ordinary reference passing to ensure consistent message ordering.
+
+Promise shortening involves peers with the following roles:
+- **Holder**: The peer holding a reference to a promise hosted by the Shortener that is being given a
+shorter promise on another peer.
+- **Shortener**: The peer hosting a promise that wishes to pass a shorter promise to the Holder.
+- **Terminator**: The peer hosting the promise that the Shortener will be passing to the Holder.
+
+**Note**: It is possible for the Holder and the Terminator to be the same peer.
```
davexunit: The same algorithm should apply in this case, too, right? Do you think it's worth stating that explicitly here?

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,53 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+**Note**: A Terminator may also play the role of Shortener for a yet-shorter promise on another peer. In that case the Shortener of the prior shortening also plays the role of Holder for the further shortening.
```
davexunit: The second sentence here makes my head spin a bit, though I understand it. Might confuse new readers. I think it's the repetition of "shorten(ing|er)"

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,53 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+When a Shortener becomes aware that a promise it is sharing to a Holder has been fulfilled with a promise
+on another peer (which could be the Holder), the Shortener sends the Holder an [`op:flush`](#opflush) message
+targeting the listener resolver for the Shortener's promise.
+
+While waiting for a response to the [`op:flush`](#opflush) message, the Shortener MUST continue to forward any
+[promise pipelining](#promise-pipelining) messages received from the Holder to the promise being shortened.
+
+The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
+new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
+a regular [Third Party Handoff](#third-party-handoffs).
```
davexunit: nit: "regular" seems extraneous here.

RidleyWrites: will remove

draft-specifications/CapTP Specification.md
RidleyWrites: I completely agree that we should have a section in the spec that clearly explains message ordering expectations. Should that be part of this PR or its own PR?

draft-specifications/CapTP Specification.md
RidleyWrites: Agreed. Will change in edit.

draft-specifications/CapTP Specification.md
RidleyWrites: Yes. Will state explicitly.

draft-specifications/CapTP Specification.md
RidleyWrites: Does this make it any better?

> In that case the peer that had the role of Shortener in the previous shortening will play the role of Holder for the further shortening

Totally see that this sentence could use improvement but short on ideas for how to do so.

draft-specifications/CapTP Specification.md
RidleyWrites: Yep. Can definitely add it explicitly.

draft-specifications/CapTP Specification.md
RidleyWrites: Sure

draft-specifications/CapTP Specification.md
davexunit: Maybe we can just file an issue, propose it be added to the milestone, and skip it here to keep the scope of the PR as it is?

RidleyWrites: Will do.

RidleyWrites: Added comment to #40 as such.

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,56 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+
+**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message the
+Shortener sent to the Holder, the Shortener becomes aware of a yet-shorter promise by receiving its own 
+[`op:flush`](#opflush) message. In that case the Shortener MUST wait until it receives the yet-shorter promise 
+and then pass *that* promise to the original Holder rather than the promise it originally intended to.
+
+## Promise shortening from the Holder's perspective
+
+When a Holder receives an [`op:flush`](#opflush) message, it creates a new 
+[promise-resolver pair](#promise-objects) and fulfills the target resolver of the [`op:flush`](#opflush) with
```
tsyesika, line 449: I think given `op:flush` technically contains two resolvers, both of which are targets in OCapN nomenclature, it'd be better to use field names (e.g. "fulfills the resolver in the flush's `to-desc` with ..." or similar).

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,56 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+the new promise. As the new promise is a local promise, this prevents any further 
+[promise pipelining](#promise-pipelining) messages being sent through this promise to the Shortener.
+
+The Holder then sends the new resolver to the resolver specified in the Shortener's [`op:flush`](#opflush)
+message.
```
tsyesika, line 454: It might be good to mention this by field name too, to make it clear since there are a lot of resolvers involved here.

draft-specifications/CapTP Specification.md
```diff
@@ -698,6 +749,30 @@ with another promise on the same peer unless that promise has been settled to
...
+`confirm-desc` MUST be a `desc:import` representing a resolver on the sender's side
+that will receive the flush confirmation containing a new resolver to send the 
+shortened promise reference.
+
+### Receiving
+
+Upon receiving the `op:flush` message, create a new [promise-resolver pair](#promise-objects).
+Fulfill the target resolver at `to-desc` with the new promise and fulfill the remote resolver
+at `confirm-desc` with the new resolver.
+
```
tsyesika, line 775: I'm wondering if it makes sense to emphasise  that, **first** the implementation fulfills the target `to-desc` (ensuring no new messages will be pipelined to the remote), **then** confirm it to the `confirm-desc`. Maybe that's overkill since it is covered in the "Promise shortening from the Holder's perspective".

tsyesika:
I left some feedback inline about trying to make it clear what promises and resolvers are being described, there's lots of them and it can sometimes be hard to follow which one is being referred to. I spent a long time thinking through this and think I might have found a couple of issues. I'll try and make some lamport diagrams to make them clear but for now I'll leave a text description.

The first thing which occurred to me is that because `op:flush` describes promises with resolvers, a promise created by specifying a answer position only isn't handled (not sure if this creates issues, but it's worth noting).

The second thing is if the `Shortener <-> Terminator` connection is slow, but the `Holder <-> Shortener` and `Holder <-> Terminator` connections are fast, I think there are cases where messages might arrive out of order still. Here are the order of events:

1. The Shortener sends their `op:flush` to the Holder
2. The Holder sends a message to the Shortener for the unresolved promise.
3. The Holder receives the flush and replies with their resolver to `confirm-desc`.
4. The Shortener receives the message for the promise and forwards it to the Terminator (slow)
5. The Shortener gets the `confirm-desc` message and does the handoff to the promise on the Terminator.
6. The Holder gets the handoff and makes a new connection to the Terminator getting the promise on that peer.
7. The Holder sends a message to the Terminator.
8. The Terminator receives the message sent in step 7.
9. The Terminator receives the forwarded message from step 4.

Unless I'm missing something those messages will arrive out of order despite the `op:flush`.

The last issue is really tricky to describe, but basically there is this paragraph:

> **Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message the
Shortener sent to the Holder, the Shortener becomes aware of a yet-shorter promise by receiving its own 
[`op:flush`](#opflush) message. In that case the Shortener MUST wait until it receives the yet-shorter promise 
and then pass *that* promise to the original Holder rather than the promise it originally intended to.

Which I think might cause a dining philosophers problem if a certain set of conditions are met as each peer is obligated to wait for the next peer. Basically if you create a loop with three peers where each has sent their `op:flush` and then reads the `op:flush` from the next peer you all end up having to wait for the next peer to respond and so they're waiting on each other causing a deadlock. If this doesn't make sense, I'm sorry, the exact situation is tricky to describe I can try and come up with a diagram.

RidleyWrites:
I will take a look at your other points yet but wanted to respond to the illustrated situation:

> Unless I'm missing something those messages will arrive out of order despite the op:flush.

Assuming I'm understanding the case as you've laid it out (correct me if wrong), the deposit-gift MUST arrive at the Terminator after any messages that were forwarded by the Shortener because of the FIFO connection between peers. 

So the holder may send messages on the handoff-withdraw promise, but that promise would not be connected to the promise being handed off until after it had processed all forwarded messages.

RidleyWrites:
> Which I think might cause a dining philosophers problem if a certain set of conditions are met as each peer is obligated to wait for the next peer. Basically if you create a loop with three peers where each has sent their op:flush and then reads the op:flush from the next peer you all end up having to wait for the next peer to respond and so they're waiting on each other causing a deadlock. If this doesn't make sense, I'm sorry, the exact situation is tricky to describe I can try and come up with a diagram.

I will think more about this, but the feature in question was necessary in order to guarantee the message order. Otherwise this algorithm does not work. I think it was @kumavis who figured that out (https://github.com/ocapn/ocapn/issues/11#issuecomment-4608367438).

RidleyWrites:
> The first thing which occurred to me is that because op:flush describes promises with resolvers, a promise created by specifying a answer position only isn't handled (not sure if this creates issues, but it's worth noting).

This is where the proposals at the bottom of #167 come in. Either of those proposals provides the means of always having a resolver for this purpose.

RidleyWrites:
> Basically if you create a loop with three peers where each has sent their op:flush and then reads the op:flush from the next peer you all end up having to wait for the next peer to respond and so they're waiting on each other causing a deadlock.

Unfortunately I failed to consider cycle scenarios. I think you just made a very good find. 

Correct me if I'm wrong but it seems the issue can be summarized in this two-party scenario:

1. Alice has p1/r1 and Bob has p2/r2
2. Alice shares p1 with Bob
3. Bob shares p2 with Alice
4. Concurrently, Alice shortens p1 to p2 and Bob shortens p2 to p1
4.1. Alice and Bob concurrently send each other op:flush
4.2. Both Alice and Bob can't respond to the flush they receive because they are waiting for a response on the flush they have sent

RidleyWrites:
I propose that for the time being, further discussion be moved back to https://github.com/ocapn/ocapn/issues/11#issuecomment-5040928304 where I have noted the cycle issue discovered by @tsyesika.

As this PR was specifically about the shortener-initiated approach, I am going to take the liberty of closing it for now. Someone tell me if that's not okay. We can re-open it later if we decide to alter the shortener-initiated algorithm.

RidleyWrites:
After much consideration I have added a commit that removes the previous cycle issue while still preserving message order even in 4+ party cases. I really hope we've got it this time.

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,61 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
+new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
+a [Third Party Handoff](#third-party-handoffs).
+
+**Important**: It is possible that while waiting for a response to the [`op:flush`](#opflush) message the
+Shortener sent to the Holder, the Shortener becomes aware of a yet-shorter promise on a second Shortener by 
+receiving its own [`op:flush`](#opflush) message. In that case the first Shortener MUST continue to forward 
+any messages from the Holder to the second Shortener until the first Shortener receives the flush response 
+from the Holder. Upon receiving the flush response from the Holder, the first Shortener MUST pass the
+reference to the next Shortener, even if it has access to an even shorter promise.
```
tsyesika: I find this a little unclear. I suspect you must pass the reference to the current closest promise you have access to on resolution?  (e.g. if you're mid-`op:flush` you should just pass the next promise regardless). I am assuming if the `op:flush` the peer receives has completed and you actually have access to an even closer promise you should supply that one.

tsyesika (approved):
I had one comment about some wording I found a bit unclear, but generally I think this looks good and I don't think it has the same cycle issues if I'm thinking about this right.

draft-specifications/CapTP Specification.md
RidleyWrites: This is covered now in #281

draft-specifications/CapTP Specification.md
RidleyWrites: I was afraid this would be unclear.

I'm not entirely sure I understand the question but I will try and explain by example to ensure we both share understanding on the intention and then hopefully we can come up with a way to clarify the wording.

If Bob sends a flush to Alice to begin shortening to a promise on Carol, Bob MUST always complete that shortening with the promise on Carol regardless of whether Bob receives a flush from Carol during that time. If he does not, then our message guarantees are broken.

If Carol sends a flush to Bob to begin shortening to a promise on Derek and Bob receives the promise on Derek BEFORE Bob sends a flush to Alice, then and only then can Bob complete the shortening with the promise on Derek.

Any suggestions for how to edit the spec text to reflect this?

tsyesika: I think since we've named the peer "second Shortener", maybe we should carry that through to describing the promise.

> "the first Shortener MUST pass the reference to the second Shortener's promise, even if it has access to a yet-shorter promise".

It still think it's going to be tricky to understand, there's just a lot of peers and promises being described. I'm struggling to come up with anything better though.

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,61 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
 
+# [Promise Shortening](#promise-shortening)
+
+When a promise is fulfilled with a promise on another peer, a layer of indirection is removed, thus 
+"shortening" the promise. Extra steps must be taken beyond ordinary reference passing to ensure 
+consistent message ordering.
+
+Promise shortening involves peers with the following roles:
+- **Holder**: The peer holding a reference to a promise hosted by the Shortener that is being given a
+shorter promise on another peer.
```
davexunit: I think simplifying this to just "The peer holding a reference to a promise hosted by the Shortener" is sufficient. I was getting confused reading this but leaving off the end of the sentence and re-reading the following definitions cleared things up for me.

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,61 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+in that case as it does when the Holder and Terminator are separate peers.
+**Note**: A Terminator may also play the role of Shortener for a yet-shorter promise on another peer. 
+In that case the Shortener of the prior shortening also plays the role of Holder for the further shortening.
+
+
+## Promise shortening from the Shortener's perspective
+
+When a Shortener becomes aware that a promise it is sharing to a Holder has been fulfilled with a promise
+on another peer (which could be the Holder), the Shortener sends the Holder an [`op:flush`](#opflush) message
+targeting the listener resolver for the Shortener's promise.
```
davexunit: Is "listener" redundant here? Or should it be "listener's resolver"?

draft-specifications/CapTP Specification.md
```diff
@@ -406,6 +407,61 @@ The Receiver should be the only party able to withdraw the gift left by the Gift
...
+
+While waiting for a response to the [`op:flush`](#opflush) message, the Shortener MUST continue to forward any
+[promise pipelining](#promise-pipelining) messages received from the Holder to the promise being shortened.
+
+The response from the Holder will contain a reference to a new resolver. The Shortener proceeds to fulfill the
+new resolver with the shorter promise. If the Holder and Terminator are not the same peer, this will involve
+a [Third Party Handoff](#third-party-handoffs).
```
davexunit, line 438: If the Terminator and Shortener are the same peer then a handoff is not needed, right?

RidleyWrites: If the terminator and shortener are the same peer then promise shortening should not be performed over OCapN.

draft-specifications/CapTP Specification.md
RidleyWrites: "listener's resolver" is better, thanks

draft-specifications/CapTP Specification.md
RidleyWrites: Sure, will edit

draft-specifications/CapTP Specification.md
RidleyWrites: Edited as suggested

PR #282, opened 2026-08-03
Support listening for promise shortening without result

### LINKED ISSUE #167: Proposal: Single op:deliver message type with resolution subscription via `op:listen`

RidleyWrites:
Previously we were able to listen to a promise or not listen to a promise. With this PR, we can listen only for promise shortening updates or we can listen for both promise shortening updates and the settled result of the promise.

Unfortunately despite this being a fairly small code change, the implementer's guide needed significant editing to accommodate it. I did my best but am very open to help refining it.

Resolves #167


jar398:
The group was favorable to this at august meeting.  Waiting for PR review.

draft-specifications/CapTP Specification.md
```diff
@@ -612,34 +617,26 @@ important to notify the other side with a [`op:gc-answer`](#opgc-answer)
...
+exported at the answer position specified by `answer-pos`. This promise MUST 
+resolve to the result the object returns. Messages sent to this promise MUST be 
+delivered to the object when the promise resolves (unless the promise breaks). 
+This promise should remain available until the [`op:gc-answer`](#opgc-answer) 
+message is received. 
 
-If `resolve-me-desc` is non-false, the result returned by the target object
-should be delivered to the specified resolver object when available.
+If the `answer-pos` is false, then promise pipelining is not used and the 
+result of the message does not need to be held.
```
tsyesika, line 639: Small nit, but if it's false promise pipelining or any regular listening for the result isn't used. It could be written:

> If the `answer-pos` is false, then the sender is not interested in the result, therefore it does not need to be exported or held.

implementation-guide/Implementation Guide.md
```diff
@@ -203,104 +201,122 @@ Okay, looks great. Alisha sends that... but she doesn't hear anything back. That
...
+The `answer-pos` contains a unique (to the CapTP session) positive integer selected by the sender to describe the promise being created. The created promise can then later be referenced using a special "answer descriptor", `desc:answer`.
+
+Alisha wants to support this mechanism so when a CapTP session is initialized she creates two further tables to facilitate this:
+
+- Questions: This is to keep track of a mapping between a remote promise descriptor and the answer position
+- Answers: This is used to keep track of position to local promises for received `op:deliver`, `op:get`, `op:index` and `op:untag` calls.
+
+Alisha implements a function which like before takes a remote object reference and a list of arguments, except this time it'll return a promise. Shhe creates a promise pair, setting the resolver aside for now, and selects an unused `answer-pos` to represent the promise. She stores the promise at the chosen `answer-pos` in the questions table. She then creates and sends an `op:deliver` with the chosen `answer-pos` instead of the `false` she used earlier. Finally, Alisha returns her internal representation for this promise back to the caller.
```
tsyesika, line 224: "Shhe" typo

tsyesika (commented):
I've left a few small comments. Generally speaking this looks good if we decide to go this direction. When I was looking at some of these examples it made me step back and think about the overall design a bit more, some of my thoughts I shared with Ridley. I wonder if we want to discuss this more at an implementor's meeting before merging.

Thanks for working on this, it's tricky to try and incorporate.

RidleyWrites:
@tsyesika I agree I think it would be good to discuss your concerns at an implementers' meeting before merging.

Issue #283, opened 2026-08-05
Inclusion of maps and sets in the data model


RidleyWrites:
Currently there are no sets and no maps that are not string-keyed in the data model. I think this is an oversight and that we should correct it.

I propose that implementers be expected to use whatever tools are available in their language to ensure that an OCapN Set or OCapN Map is equal to itself on round-trip. That means comparison of contents, not of identity. The data model already requires this for Lists (though I think we need to make this requirement more obvious as I missed it for a long time).

I don't know how this would be handled in Scheme-land but my understanding of Endo-land would suggest that an OCapN Set would be decoded as a CopySet and an OCapN Map would be decoded as a CopyMap. I think OCapN List is already decoded as CopyList if I understand correctly. My implementation currently works like this.

While it is possible to build collections on top of Tagged, it feels to me like a major oversight to not include sets and maps in the data model because users are going to expect them. If not required by the data model, we are only moving the problem into user space. By including sets and maps in the data model we can also require canonicalization which is less likely to be consistent if these collections are pushed to user space.

davexunit:
I support this. As Richard Gibson said at the last implementers meeting, once you have maps you have sets, so coming to consensus on maps is the main issue at hand. There's been disagreement over the set of valid keys, which is why it is over-constrained (imo) to only strings right now. I think any value should be acceptable as a key as it is both simple and flexible. The OCapN spec should contain a definition of equality in which primitive and compound values are compared by structure and references are compared by identity.

dckc:
> I think this is an oversight

it is not an oversight. It was considered in depth in the May 2023 meeting.

 - https://github.com/ocapn/ocapn/issues/43
   - [OCapN & Agoric: Layers and Orders](https://www.youtube.com/watch?v=htS-2gvY3Cs) - 47min presentation recording

If there's something we didn't consider at the time, please let us know.

davexunit:
I think @RidleyWrites has pointed out some things that were not considered years ago, such as how leaving out these compound types pushes the canonicalization burden to userspace.

dckc:
Leaving the keyEQ burden to the patterns layer is discussed [around 12:00 in the presentation](https://youtu.be/htS-2gvY3Cs?si=TlxrXzDuV4ZR9jRH&t=725).

_p.s. perhaps I haven't kept up with where canonicalization is used in the protocol._

RidleyWrites:
[...]

Issue #172, opened 2025-05-15
Reusability of passable data


kriskowal:
Related: #168, #169 

Agoric’s experience with “passable values” (values within OCapN’s model, or “cap data”) suggests that there will be two common arrangements where the byte-wise representation of a passable value can get reused without modification. These motivate a desirable complication of OCapN’s CapTP.

In this simple case, we have a “kernel” that coordinates messages among a set of local vats. The kernel vat coordinates the clists for each child vat and does not need to reïfy passable values locally. The kernel is responsible for ensuring that a target passed from vat to vat always arrives with a unique and comparable identity (“grant matching”). If the slot numbers for each respective target and promise in the body of a message are separate, the kernel can pass the body unmodified and does not need to suffer a round-trip through marshaling.

```mermaid
graph LR
  KERNEL <--> VAT1
  KERNEL <--> VAT2
  KERNEL <--> VAT3
```

In a more complex case, the “kernel” introduces a “comms vat” to communicate with other peers. The “comms vat” maps slot numbers from the various open OCapN sessions on the wire, to its own slot numbers presented to other vats through the kernel. This is another opportunity to avoid a round-trip through serialization for the body of passable data.

```mermaid
graph LR
  KERNEL <--> VAT1
  KERNEL <--> VAT2
  KERNEL <--> VAT3
  KERNEL <--> COMMSVAT
  COMMSVAT <--> AVERY
  COMMSVAT <--> BOBBY
  COMMSVAT <--> CHARLIE
```

OCapN as written today does not decouple the C-list positions of promises and targets in a message from the structure of the value. Consider a value that contains a couple promises, a couple targets, and an error.

```
[ <import-object -10> <import-object 2> <import-promise 3> <import-promise 50> <error "TypeError">]
```

I propose an alternate representation. In this representation, the body is a ByteArray that contains Syrup. The ByteArray envelope gives the parent message a length prefix so it can fast-forward beyond the body without having to match nested Structs, Lists, and Records. The remaining values in the record are parallel arrays with out-of-body information about respective entities in-the-body.

```
<'passable
  :[ <target> <target> <promise> <promise> <error "TypeError">]:  ; Body
  [ -10 2 ]                                                       ; Respective targets for position
  [ 3 50 ]                                                        ; Respective promises
  [ <'error swissnum "no no"> ]                                   ; Respective errors
>
```

[...]

Issue #142, opened 2024-11-19
Error type and passage invariants


kriskowal:
We have not yet settled any details on an OCapN Error type except that there ought to be one. Some of the parameters of the design include:

- Whether errors have pass-invariant equality. That is, if an error can be sent to another peer, returned by that peer, and for the sent and received error to be equal, and unequal to any other value.
- Whether errors are associated with a pass-invariant equality. That is, supposing errors are Struct-like, then its reference value might uniquely associate it with a token that does have pass-invariant equality.
- If struct-like, whether any particular fields of Errors are reserved, forbidden, or convey meaning.
- If struct-like, whether `stack` is forbidden, or if JavaScript is expected to carry reïfy it differently than other names. For example, on a `Symbol.for('stack')` to distinguish it from the `.stack`.
- If struct-like, whether `name` is forbidden, limited to particular String names, whether JavaScript muxes the type on those names to the corresponding JavaScript types, or similarly diverted to `Symbol.for('name')`.
- If struct-like, whether to carry other arbitrary fields.

An error should be suitable for both conveying “semantic” information in-band and also convey an association with “diagnostique” out-of-band information, to a party that has a debugging capability. To that end, we need a design that is both struct-like, in that it can convey arbitrary data in-band, and also reference/capability-like, in that it can be used as a token for diagnostics.

Having the Error value be primarily struct-like for purposes of conveying in-band semantic information and providing an API to asynchronously obtain the associated diagnostique token has a compelling property: The diagnostique token can be a promise for a target obtained from a third-party debugger, such that these are constructed lazily and allow the holder to connect directly to the debugger.

erights:
See also https://github.com/endojs/endo/blob/master/packages/ses/src/error/README.md for lots of info on endo's approach to errors. Not all of this will be relevant to ocapn, and some of this is likely already stale for endo. But it is important background for the conversation we need to have here.

kriskowal:
I want to propose that Errors be modeled as a strange hybrid, that they are like structs in that they carry data, and also not-like-structs in that they would have pass-invariant equality, so you can 3pho errors with a debugger vat. Because error objects vary wildly between languages, where each language is likely to reserve certain names, we may need to take some care to maintain  the round-tripping invariant.

davexunit:
[...]

Issue #66, opened 2023-06-17
S-expressions in keys and signatures


dpwiz:
The spec mandates using plain sequences here, referring to "gcrypt".
This looks like some 3rd party presentation leaking.

The named arguments do sure look like dictionary:
```
<public-key ecc {curve: Ed25519, flags: eddsa, q: ..., s: ...}>
```
Why not use them instead?

RidleyWrites:
I would like if both the 'public-key and 'sig-val would be represented in some way as Syrup records, since that is how custom data types are handled in Syrup. It would be more consistent and one less thing for a new implementer to worry about.

RidleyWrites:
We had some discussion on this issue at today's implementers' meeting.

I think it was @cwebber who suggested that instead of the current s-expressions for keys and signatures, we could move the crypto-suite info elsewhere and just use the key/signature bytes where we are currently using the s-expressions. Or we could assign `Public Key` and `Signature` types in the new encoding but I don't know that that's necessary.

The primary suggestion which I like is to include a crypto-suite version in the start-session info in addition to the captp-version. 

There was also a question of whether the crypto-suite info would also be needed in certificates. I lean towards no but am not confident in that opinion.

RidleyWrites:
To have something slightly more concrete to discuss, here is a possibility:

- signature field of `desc:sig-envelope` is ByteArray
- receiver-key field of `desc:handoff-give` is ByteArray
- session-pubkey field of `op:start-session` is ByteArray
- `op:start-session` contains a crypto-version String field

Unanswered questions:
- what crypto-version String to use to represent current crypto?
- do handoff certificates need to include crypto-version somehow?

RidleyWrites:
> do handoff certificates need to include crypto-version somehow?

At today's implementers' meeting we agreed to not require this.

I will work on a PR.

PR #284, opened 2026-08-19
OCapN encoding

### LINKED ISSUE #4: Verbose record labels while we figure things out, enums or short labels later
Right now, Goblins' CapTP representation (see also #3) uses very verbose records: they look like `<'op:deliver-only ...>`; every record label is a full stringy-symbol.  This is not desirable in the final version, but it makes it very easy to reason about the protocol now.  It is easy to type out a description of the protocol that is moderately human readable in the [textual representation of preserves](https://preserves.gitlab.io/preserves/preserves.html#textual-syntax).

However, when we ship, we probably want to enum'ify this... replace all of these with short bytes, strings, symbols or integers... no more than an octet in length each.  But I suggest we do this later in the design, so we can keep our brains well wrapped around things.

### LINKED ISSUE #275: Consensus on concrete encoding
The specification tentatively specifies that messages are encoded as Syrup. The specification itself is written in Preserves, as a diagnostic notation or representative format of Syrup. @gibson042 and I have sketched a CBOR #249 alternative. The plenary group has agreed to defer the topic of the concrete encoding format until thornier issues have settled, but this will need to be settled in the scope of a first consensus draft specification.

I offer the following summary of positions in [IBIS notation (issue based information system)](https://en.wikipedia.org/wiki/Issue-based_information_system):

```ibis
? What concrete format should we use for OCapN messages?
  . Syrup
    - Provides a false sense of human legibility. Many forms are human-readable which
       may lead to habitual attempts to directly inspect payloads. However, byte strings
       may contain Syrup and will appear to be Syrup to an unwary reader, which can lead
       to misinterpretation.
    - Lineage can be traced through bencode and netstring, however not widely recognized,
      few reference implementations.
      + On a backdrop of no *non-canonical* reference implementations, we can ensure
         all Syrup implementations are *canonical*.
    - Sacrifices modest compactness for
    + Modestly easier to implement than CBOR
  . CBOR
    + Relatively widely recognized
    + Supports canonical form
      - Canonical form is not the norm and implementations are likely to dissemble
        + However, OCapN does not use the entirety of Syrup or CBOR and because
           OCapN is a subset of these and implementations of OCapN will provide better baseline
           performance in general if it eschews an intermediate representation format,
           OCapN implementations are only likely to benefit from an off-the-shelf implementation
           of Syrup or CBOR in a rough sketch, and using the IR directly is likely to allow
           programs to express non-passable values anyway.
           So, for an OCapN implementation, the existence of a Syrup or CBOR codec
           does not imply one should use either above rolling an OCapN specific implementation
    + Wire debugging tools like cbor.me exist for direct inspection (where canonicalizability does not matter)
    + Sacrifices modest ease of implementation in exchange for
    - Slightly more compact than Syrup (e.g., single byte small integers and length prefixes)
  . non-differentiating concerns between Syrup and CBOR
    + Sufficient to model all OCapN passable data
      - And also some impassible data
    + Sufficient compactness
    + Sufficient ease of implementation
    + Sufficiently ergonomic representation format (Preserves or CBOR diagnostic notations),
       suitable for specifying shapes abstractly
    + Can fast forward through byte strings
      - But cannot through complex structures unless enveloped in a byte string
    + Byte strings do not suffer escaping or hotelling, in contrast to JSON
```

RidleyWrites:
Following discussions in #275 and in implementers' meetings, I have put together a working document.

The following are the issues that I believe require further discussion or review which will impact this encoding:

- [x] #66 
- [ ] #283 
- [ ] #172 
- [ ] #142 
- [ ] #277 

I also have a couple TODO questions in here to talk about at some point.

I feel like the document presentation could be improved but at least this gives us a starting point.

It is possible that work will be needed to adjust existing documents to reflect the encoding. If so I request that others chip in to help with that.

### DIFF
--- a/draft-specifications/Encoding.md
+++ b/draft-specifications/Encoding.md
@@ -235,6 +235,8 @@ tag 0x19
 
 0x24 || \<\<signed>> || \<\<ByteArray signature>>
 
+//TODO: signature format
+
 #### desc:handoff-give
 `<desc:handoff-give receiver-key exporter-location session-id gifter-side gift-id>`  
 
@@ -247,9 +249,11 @@ tag 0x19
 
 ### OCapN Operators
 #### op:start-session
-`<op:start-session captp-version crypto-version session-pubkey acceptable-location acceptable-location-sig>`  
+`<op:start-session captp-version session-pubkey acceptable-location acceptable-location-sig>`  
 
-0x30 || \<\<String captp-version>> || \<\<String crypto-version>> || \<\<ByteArray session-pubkey>> || \<\<PeerLocator acceptable-location>> || \<\<desc:sig-envelope acceptable-location-sig>>
+0x30 || \<\<String captp-version>> || \<\<session-pubkey>> || \<\<PeerLocator acceptable-location>> || \<\<desc:sig-envelope acceptable-location-sig>>
+
+//TODO: public key format
 
 #### op:abort
 `<op:abort reason>`  
--- a/draft-specifications/Encoding.md
+++ b/draft-specifications/Encoding.md
@@ -1,0 +1,309 @@
+# OCapN Encoding
+
+## Tags
+
+Every type in this encoding format is represented by a tag byte potentially followed by some content as described in [details](#details).
+
+| Tag (Hex) | Type |
+| --- | --- |
+| 00  | Null |
+| 01  | Undefined |
+| 02  | False |
+| 03  | True |
+| 04  | Double |
+| 05  | Positive integer (1 Byte) |
+| 06  | Positive integer (2 Bytes) |
+| 07  | Positive integer (4 Bytes) |
+| 08  | Positive integer (8 Bytes) |
+| 09  | Positive BigInt |
+| 0A  | Negative integer (1 Byte) |
+| 0B  | Negative integer (2 Bytes) |
+| 0C  | Negative integer (4 Bytes) |
+| 0D  | Negative integer (8 Bytes) |
+| 0E  | Negative BigInt |
+| 0F  | RESERVED |
+
+| Tag (Hex) | Type |
+| --- | --- |
+| 10  | ByteArray |
+| 11  | String |
+| 12  | Symbol |
+| 13  | Tagged |
+| 14  | List |
+| 15  | Set |
+| 16  | Map |
+| 17  | Peer Locator |
+| 18  | Sturdy Ref |
+| 19  | Error |
+| 1A-1F | RESERVED |
+
+| Tag (Hex) | Type |
+| --- | --- |
+| 20  | desc:answer |
+| 21  | desc:export |
+| 22  | desc:import-object |
+| 23  | desc:import-promise |
+| 24  | desc:sig-envelope |
+| 25  | desc:handoff-give |
+| 26  | desc:handoff-receive |
+| 27-2F | RESERVED |
+
+| Tag (Hex) | Type |
+| --- | --- |
+| 30  | op:start-session |
+| 31  | op:abort |
+| 32  | op:deliver |
+| 33  | op:listen |
+| 34  | op:gc-exports |
+| 35  | op:gc-answers |
+| 36  | op:get |
+| 37  | op:index |
+| 38  | op:untag |
+| 39  | op:flush |
+| 3A-FF | RESERVED |
+
+## Notations
+
+Two representations are displayed in order:
+
+- Text (for things such as spec writing)
+- Binary
+
+A || B denotes binary concatenation of A and B  
+\<\<something>> denotes an encoded representation of something
+
+## Varint format
+In various places this spec uses variable-length integers to denote length or size of data being transmitted.
+
+Varints that can fit in 7 bits are encoded as is with a leading 0 bit.
+
+Varints longer than 7 bits are encoded as the minimum number of bytes required to represent the integer,  
+preceded by a leading byte with value `0x80 | numBytes`.
+
+For example, the varint representation of the integer 255 would be 0x81FF.
+
+(Note that this is not how integer data is encoded. See [integer](#integer) for more.)
+
+## Details
+### Null
+`null`
+
+0x00
+
+### Undefined
+`undefined`
+
+0x01
+
+### False
+`false`
+
+0x02
+
+### True
+`true`
+
+0x03
+
+### Double
+
+`123.0` <- decimal important to differentiate from int
+
+0x04 || 8 IEEE 754 encoded bytes
+
+### Integer
+
+Integers are represented using a different tag depending on the minimum number of bytes required to encode the absolute value of the integer. For example, the integer 4 would be encoded as 0x0504 while the integer 256 would be encoded as 0x060100.
+
+Negative integers are represented by picking the appropriate negative tag byte in the range 0x0A to 0x0E. Following the tag, values are transmitted as absolute values (-n). Negative 0 is invalid.
+
+#### Positive integer (1 Byte)
+0x05 || byte
+
+#### Positive integer (2 Bytes)
+0x06 || bytes
+
+#### Positive integer (4 Bytes)
+0x07 || bytes
+
+#### Positive integer (8 Bytes)
+0x08 || bytes
+
+#### Positive integer (BigInt)
+0x09 || [varint](#varint-format) num bytes || bytes
+
+#### Negative integer (1 Byte)
+0x0A || byte
+
+#### Negative integer (2 Bytes)
+0x0B || bytes
+
+#### Negative integer (4 Bytes)
+0x0C || bytes
+
+#### Negative integer (8 Bytes)
+0x0D || bytes
+
+#### Negative integer (BigInt)
+0x0E || [varint](#varint-format) num bytes || bytes
+
+### ByteArray, String, Symbol
+tag || [varint](#varint-format) num bytes || bytes
+
+For String and Symbol, bytes is the UTF-8 encoding of the String and Symbol name, respectively.
+
+| Text | Binary |
+| --- | --- |
+| 0xABC123 | 0x10 \|\| 0x06 \|\| 0xABC123 |
+| "my-string" | 0x11 \|\| 0x09 \|\| bytes |
+| 'my-symbol | 0x12 \|\| 0x09 \|\| bytes |
+
+### Tagged
+`<name value>`  
+
+0x13 || \<\<String name>> || \<\<value>>
+
+//TODO: Symbol name?
+
+### List
+`[e1, e2, e3]`  
+
+0x14 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>
+
+### Set
+`{e1, e2, e3}`  
+
+0x15 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>
+
+All elements must be unique and written in order sorted lexicographically by their representations.
+
+### Map
+`{k1: v1, k2: v2, ... kN: vN}`  
+
+0x16 || [varint](#varint-format) N elements || \<\<k1>> || \<\<v1>> || \<\<k2>> || \<\<v2>> || ... || \<\<kN>> || \<\<vN>>
+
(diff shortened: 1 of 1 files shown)


draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+| 05  | Positive integer (1 Byte) |
+| 06  | Positive integer (2 Bytes) |
+| 07  | Positive integer (4 Bytes) |
+| 08  | Positive integer (8 Bytes) |
+| 09  | Positive BigInt |
+| 0A  | Negative integer (1 Byte) |
+| 0B  | Negative integer (2 Bytes) |
+| 0C  | Negative integer (4 Bytes) |
+| 0D  | Negative integer (8 Bytes) |
+| 0E  | Negative BigInt |
```
gibson042: Agreement with https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md :
```suggestion
| 02  | Boolean false |
| 03  | Boolean true |
| 04  | Float64 |
| 05  | positive Integer (1 bytes) |
| 06  | positive Integer (2 bytes) |
| 07  | positive Integer (4 bytes) |
| 08  | positive Integer (8 bytes) |
| 09  | positive Integer (>8 bytes) |
| 0A  | negative Integer (1 bytes) |
| 0B  | negative Integer (2 bytes) |
| 0C  | negative Integer (4 bytes) |
| 0D  | negative Integer (8 bytes) |
| 0E  | negative Integer (>8 bytes) |
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+| --- | --- |
+| 10  | ByteArray |
+| 11  | String |
+| 12  | Symbol |
+| 13  | Tagged |
+| 14  | List |
+| 15  | Set |
+| 16  | Map |
+| 17  | Peer Locator |
+| 18  | Sturdy Ref |
```
gibson042:
```suggestion
| 18  | sturdyref locator |
```
[?](https://github.com/ocapn/ocapn/blob/main/draft-specifications/Locators.md#sturdyref-locator)

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+Two representations are displayed in order:
+
+- Text (for things such as spec writing)
+- Binary
+
+A || B denotes binary concatenation of A and B  
+\<\<something>> denotes an encoded representation of something
+
+## Varint format
+In various places this spec uses variable-length integers to denote length or size of data being transmitted.
```
gibson042:
```suggestion
In various places this spec uses variable-length integers (varints) to denote length or size of data being transmitted.
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+Varints that can fit in 7 bits are encoded as is with a leading 0 bit.
+
+Varints longer than 7 bits are encoded as the minimum number of bytes required to represent the integer,  
+preceded by a leading byte with value `0x80 | numBytes`.
+
+For example, the varint representation of the integer 255 would be 0x81FF.
+
+(Note that this is not how integer data is encoded. See [integer](#integer) for more.)
```
gibson042:
```suggestion
(Note that this is not how Integer data is encoded. See [Integer](#integer) for more.)
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+`false`
+
+0x02
+
+### True
+`true`
+
+0x03
+
+### Double
```
gibson042:
```suggestion
### Float64
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+`123.0` <- decimal important to differentiate from int
```
gibson042:
```suggestion
`123.0` <- decimal important to differentiate from Integer
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+0x0E || [varint](#varint-format) num bytes || bytes
+
+### ByteArray, String, Symbol
+tag || [varint](#varint-format) num bytes || bytes
+
+For String and Symbol, bytes is the UTF-8 encoding of the String and Symbol name, respectively.
+
+| Text | Binary |
+| --- | --- |
+| 0xABC123 | 0x10 \|\| 0x06 \|\| 0xABC123 |
```
gibson042: Is this `0x06` a typo?
```suggestion
| 0xABC123 | 0x10 \|\| 0x03 \|\| 0xAB_C1_23 |
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Double
+
+`123.0` <- decimal important to differentiate from int
+
+0x04 || 8 IEEE 754 encoded bytes
+
+### Integer
+
+Integers are represented using a different tag depending on the minimum number of bytes required to encode the absolute value of the integer. For example, the integer 4 would be encoded as 0x0504 while the integer 256 would be encoded as 0x060100.
```
gibson042:
```suggestion
### Integer

`42`

Integers are represented using a different tag depending on the minimum number of bytes required to encode the absolute value of the integer. For example, the integer 4 would be encoded as 0x0504 while the integer 256 would be encoded as 0x060100.
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+| Text | Binary |
+| --- | --- |
+| 0xABC123 | 0x10 \|\| 0x06 \|\| 0xABC123 |
+| "my-string" | 0x11 \|\| 0x09 \|\| bytes |
+| 'my-symbol | 0x12 \|\| 0x09 \|\| bytes |
+
+### Tagged
+`<name value>`  
+
+0x13 || \<\<String name>> || \<\<value>>
```
gibson042: It is important to show the presence (or respectively absence) of the tag byte for a Tagged's name.
```suggestion
0x13 || \<\<String name>> || \<\<value>>

example: `<foo true>` is 0x13 || 0x11_03_666F6F || 0x03
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+//TODO: Symbol name?
+
+### List
+`[e1, e2, e3]`  
+
+0x14 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>
```
gibson042:
```suggestion
0x14 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>

example: `[true, false]` is 0x14 || 0x02 || 0x03 || 0x02
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Set
+`{e1, e2, e3}`  
+
+0x15 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>
```
gibson042:
```suggestion
0x15 || [varint](#varint-format) N elements || \<\<e1>> || \<\<e2>> || ... || \<\<eN>>

example: `{false, true}` is 0x15 || 0x02 || 0x02 || 0x03
```

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+All elements must be unique and written in order sorted lexicographically by their representations.
+
+### Map
+`{k1: v1, k2: v2, ... kN: vN}`  
+
+0x16 || [varint](#varint-format) N elements || \<\<k1>> || \<\<v1>> || \<\<k2>> || \<\<v2>> || ... || \<\<kN>> || \<\<vN>>
```
gibson042:
```suggestion
0x16 || [varint](#varint-format) N elements || \<\<k1>> || \<\<v1>> || \<\<k2>> || \<\<v2>> || ... || \<\<kN>> || \<\<vN>>

example: `{"": true, ': true}` is 0x16 || 0x02 || 0x11_00 || 0x03 || 0x12_00 || 0x03
```
Hmm, this suggests that `'my-symbol` is not a good presentation format for symbols.

gibson042 (commented):
Preliminary comments.

draft-specifications/Encoding.md
RidleyWrites: Oops good catch. I think my head thought I was looking at UTF-8 for some reason.

RidleyWrites:
Thanks @gibson042 for the editorial pass!

draft-specifications/Encoding.md
RidleyWrites: Can you clarify what the problem with `'my-symbol` is? Is it that you are using a symbol with an empty name for this example? Or is it a Markdown issue? I am fine to change formats if needed.

Which raises the question: Do we need to restrict what symbol names are allowed to be? Because right now I think they're just any string according to spec but IIRC Guile would not permit an empty name or a name with spaces.

I can accommodate whatever since I wasn't able to use Dart's built-in Symbol to represent OCapN symbols.

gibson042: > Can you clarify what the problem with `'my-symbol` is? Is it that you are using a symbol with an empty name for this example?

Yes, but because the proposed presentation format for Symbol is unterminated, even the following `:` is a problem... e.g., does `{':':0}` map the empty-named Symbol to the Symbol named `:0`, or does it map the Symbol named `:'` to the Integer 0?

[Notation](https://github.com/ocapn/ocapn/blob/main/draft-specifications/Notation.md#symbol) gets around this by explicitly constraining the presentation format in exclusion of valid Symbols whose name contains anything other than a–z/A–Z/0–9/`-`/`:` or doesn't start with an ASCII letter... maybe this document should just follow suit?

> Which raises the question: Do we need to restrict what symbol names are allowed to be? Because right now I think they're just any string according to spec but IIRC Guile would not permit an empty name or a name with spaces.

https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md#symbol is very clear about Symbol and String having identical content domains (unrestricted beyond Unicode encodability). But I'm not even clear on the arguments for having two kinds of string in a messaging protocol at all, so :shrug:.

draft-specifications/Encoding.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Varint format
+In various places this spec uses variable-length integers to denote length or size of data being transmitted.
+
+Varints that can fit in 7 bits are encoded as is with a leading 0 bit.
+
+Varints longer than 7 bits are encoded as the minimum number of bytes required to represent the integer,  
+preceded by a leading byte with value `0x80 | numBytes`.
+
+For example, the varint representation of the integer 255 would be 0x81FF.
```
jobol: Really? Isn't it 0x817F instead of 0x81FF? There should be a way to detect the end no?

gibson042: 255 is 0xFF, which needs one byte, so the first byte of its encoding is 0x80 | 1 = 0x81 and indicates that a single byte follows.

| Decimal | Hexadecimal | Byte Count | Encoding |
|--------|--------|--------|--------|
| 0 | 0x0 | 1 | 0x00 |
| 127 | 0x7F | 1 | 0x7F |
| 128 | 0x80 | 1 | 0x81_80 |
| 255 | 0xFF | 1 | 0x81_FF |
| 256 | 0x100 | 2 | 0x82_0100 |

draft-specifications/Encoding.md
RidleyWrites: I see.

I don't get much out of having symbols in the protocol. Dart Symbols are valuable but can only be created at compile time and are thus useless for the purposes of this protocol.

Given that removing them from the protocol does not seem to be on the table, I would like one of the Spritely people to comment on whether the currently documented lack of restrictions on symbol names is correct from their perspective.

RidleyWrites: I confirmed today with @tsyesika that I was wrong and Symbols are not as restricted in Guile as I thought.

In the implementers' meeting we agreed that for the time being it's okay to take the same path as we did in Notation and explore a more complicated symbol text representation later if we feel it to be necessary.

PR #288, opened 2026-09-01
Move crypto-version from public key and signature to op:start-session

### LINKED ISSUE #66: S-expressions in keys and signatures

RidleyWrites:
Resolves https://github.com/ocapn/ocapn/issues/66

### DIFF
--- a/draft-specifications/CapTP Specification.md
+++ b/draft-specifications/CapTP Specification.md
@@ -246,26 +246,13 @@ side as they become fulfilled and send the car noise back at the end.
 
 # [Cryptography](#cryptography)
 
-Each party within a CapTP session has their own per session key pair which is
-used for signing certain structures for example in the [third party handoffs
-section](#third-party-handoffs). These key pair values are generated EdDSA with
-a SHA512 hash.
-
-**NOTE:** These representations are considered temporary and we are anticipating
-replacing them, probably with record-based representations.
+Each party within a CapTP session has their own Ed25519 per session key pair which is
+used for signing certain structures - for example in the [third party handoffs
+section](#third-party-handoffs). 
 
 ## [Public Key](#public-key)
 
-Public keys are formatted based on gcrypt's s-expression format, using EdDSA
-public keys and the SHA512 hash algorithm. The EdDSA public keys are based on
-the Ed25519 elliptic curve. The public key is formatted as follows:
-
-```text
-['public-key ['ecc ['curve 'Ed25519] ['flags 'eddsa] ['q q_value]]]
-```
-
-In the above format, the `q_value` is a [ByteArray][Model-ByteArray] value of
-32 bytes, representing the public key.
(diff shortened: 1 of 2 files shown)


RidleyWrites:
Open to alternative suggestions for the crypto-version string

draft-specifications/CapTP Specification.md
```diff
@@ -514,6 +494,8 @@ instead of creating a new one.
 
 The `captp-version` MUST be `1.0`.
 
+The `crypto-version` MUST be `Ed25519`.
```
tsyesika: I think we might want to include the hashing algorithm too, e.g. "Ed25519_SHA256".

implementation-guide/Implementation Guide.md
```diff
@@ -59,9 +59,10 @@ The first messages exchanged over a CapTP session are to initialize the CapTP se
 
 ```
 <op:start-session captp-version             ; String value
-                  session-pubkey            ; CapTP public key value
+                  crypto-version            ; String value
+                  session-pubkey            ; CapTP public key value (Byte Array)
                   acceptable-location       ; OCapN Reference type
-                  acceptable-location-sig>  ; CapTP signature
+                  acceptable-location-sig>  ; CapTP signature (Byte Array)
```
tsyesika: Small nit, "ByteArray" (no space).

tsyesika (approved):
Looks great, thanks for doing this. The only real feedback (other than a tiny stylistic nit) is that we should probably include the hashing algorithm in alongside the signing cryptography.

draft-specifications/CapTP Specification.md
RidleyWrites: The hashing algorithm used to make session IDs? Do we use hashing elsewhere?

tsyesika: Right, it's not used elsewhere from memory, but obviously the session ID and peer ID which both use it have it within handoff certs. Maybe it's overkill to have it but it felt to me like it's part of the crypto suite we use (maybe I've just been in the Noise world too much recently).

RidleyWrites: Fine with me

RidleyWrites:
> Looks great, thanks for doing this. The only real feedback (other than a tiny stylistic nit) is that we should probably include the hashing algorithm in alongside the signing cryptography.

Done

RidleyWrites:
Reached consensus by group. Awaiting copy edit from @kriskowal  before merge.

kriskowal (approved):
I frankly love this. Thank you.

